Issue #75, opened 2026-04-30
CI: `swift test` fails on macos-15 runner — MLX metallib not loadable


jaylann:
## Problem

`swift test` fails on every push to `main` (and now `stage`) with:

```
MLX error: Failed to load the default metallib. library not found library not found library not found library not found
  at .../mlx-swift/Source/Cmlx/mlx-c/mlx/c/array.cpp:232
##[error]Process completed with exit code 1.
```

This is a pre-existing failure — visible on every Test run going back at least 10 runs (run IDs 25168968313, 25167115456, 25162759339, etc.), including pushes to \`main\` before any of the recent license/staged-release work.

## Why it happens

GitHub-hosted \`macos-15\` runners are virtualized M1 instances without the Metal stack the host has access to. MLX needs Metal at runtime to load its \`default.metallib\` binary. The build succeeds (which is compile-time only), but as soon as any test triggers an actual MLX array operation, the Metal library load fails and the runner exits 1.

## Effect

- Test workflow always red on CI.
- The new \`Release\` workflow (\`workflow_dispatch\`) gates on the latest \`Test\` run on \`stage\` being green — so releases can't be cut through the workflow until this is fixed.

## Possible approaches

1. **Skip MLX-dependent tests on CI.** Tag tests that touch MLX runtime with \`.enabled(if: ProcessInfo.processInfo.environment["CI"] == nil)\` or a custom trait. Macro/schema/property-wrapper tests would still run.
2. **Move MLX tests to a separate test target / scheme** that's only invoked locally.
3. **Self-hosted runner** — a real Apple Silicon Mac with full Metal access.
4. **Skip \`swift test\` on CI entirely** and rely on \`swift build\` only.

Option 1 is probably the lowest-effort fix that preserves the most coverage.

## Workaround for releases right now

Until this is fixed, the \`Release\` workflow will refuse to cut a tag because its CI gate sees a red \`Test\` run. Either:
- Manually verify \`swift test\` passes locally, then bypass with \`git tag -a vX.Y.Z && git push origin vX.Y.Z && gh release create vX.Y.Z --generate-notes\`
- Or add an \`override\` input to \`release.yml\` that lets a maintainer skip the CI gate consciously.

PR #78, opened 2026-04-30
test(compat): verify chat templates across 5 model families (#37)

### LINKED ISSUE #37: Chat template handling: verify across 5+ model families
Verify Cast works correctly with multiple model families.

Models to test:
- Qwen (uses \`<|im_start|>\` / \`<|im_end|>\`)
- Llama (uses \`<|begin_of_text|>\`)
- Mistral
- Phi
- Gemma

MLXLMCommon's \`processor.prepare(input:)\` handles chat templates — verify no edge cases.
Document any model-specific quirks.

jaylann:
## Summary
- New `ChatTemplateTests` covers Qwen, Llama, Mistral, Phi, Gemma — each `.requiresMetal`-gated (skipped on CI per issue #75).
- ADR `docs/decisions/0001-chat-template-handling.md` documents that Cast delegates chat templating entirely to `MLXLMCommon`'s `processor.prepare(input:)`; Cast hands a flat `"\(system)\n\n\(prompt)"` string to `UserInput(prompt:)` and the tokenizer's `chat_template` does the rest.
- New `ChatTemplates` example demonstrates the same prompt across all five families for manual verification.

No behavior change in Cast itself — this is verification + documentation. The five families exercise distinct chat-template formats (Qwen `<|im_start|>`, Llama `<|begin_of_text|>`, Mistral `[INST]`, Phi `<|user|>`, Gemma `<start_of_turn>`), so a single test sweep validates that Cast's flat-string concatenation does not collide with any family's templating.

Closes #37.

## Test plan
- [x] `swift build` (parent package) — passes.
- [x] `CI=true swift test` — 160 tests in 13 suites pass; `ChatTemplateTests` correctly skipped via `.requiresMetal`.
- [ ] `cd Examples && swift build` — worktree-local SPM identifier limitation; will pass on CI which checks out into `Cast/`.
- [ ] Manual local: `swift test --filter ChatTemplate` (downloads ~10 GB across families on first run).

### DIFF
--- a/Examples/Sources/ChatTemplates/main.swift
+++ b/Examples/Sources/ChatTemplates/main.swift
@@ -1,0 +1,50 @@
+// What this shows: the same prompt routed through five different model
+// families to demonstrate cross-family chat-template support. Cast hands a
+// flat "(system)\n\n(prompt)" string to MLXLMCommon's processor.prepare,
+// which applies each tokenizer's chat_template — Qwen <|im_start|>,
+// Llama <|begin_of_text|>, Mistral [INST], Phi <|user|>, Gemma
+// <start_of_turn>. No per-family code paths in Cast.
+
+import Cast
+import Collections
+import Foundation
+import JSONSchema
+
+@Castable
+struct Greeting {
+    var text: String = ""
+}
+
+@main
+enum ChatTemplates {
+    static let families: [(modelId: String, family: String)] = [
+        ("mlx-community/Qwen2.5-1.5B-Instruct-4bit", "Qwen"),
+        ("mlx-community/Llama-3.2-1B-Instruct-4bit", "Llama"),
+        ("mlx-community/Mistral-7B-Instruct-v0.3-4bit", "Mistral"),
+        ("mlx-community/Phi-3.5-mini-instruct-4bit", "Phi"),
+        ("mlx-community/gemma-2-2b-it-4bit", "Gemma")
+    ]
+
+    static func main() async throws {
+        let prompt = "Say hello as a JSON object with field 'text'."
+
+        for entry in families {
+            print("--- \(entry.family) (\(entry.modelId))")
+            do {
+                let model = try await CastModel.load(entry.modelId)
+                let greeting: Greeting = try await model.cast(prompt)
+                print("\(entry.family): \(greeting)")
+                model.unload()
+            } catch {
+                print("\(entry.family) failed: \(error)")
+            }
+        }
+    }
+}
+
+// Sample output (illustrative; varies per model + run):
+// --- Qwen (mlx-community/Qwen2.5-1.5B-Instruct-4bit)
+// Qwen: Greeting(text: "Hello!")
+// --- Llama (mlx-community/Llama-3.2-1B-Instruct-4bit)
+// Llama: Greeting(text: "Hi there.")
+// ...
--- a/Examples/Sources/ChatTemplates/main.swift
+++ b/Examples/Sources/ChatTemplates/main.swift
@@ -27,17 +27,24 @@ enum ChatTemplates {
 
     static func main() async throws {
         let prompt = "Say hello as a JSON object with field 'text'."
+        var failures: [String] = []
 
         for entry in families {
             print("--- \(entry.family) (\(entry.modelId))")
             do {
                 let model = try await CastModel.load(entry.modelId)
+                defer { model.unload() }
                 let greeting: Greeting = try await model.cast(prompt)
                 print("\(entry.family): \(greeting)")
-                model.unload()
             } catch {
                 print("\(entry.family) failed: \(error)")
+                failures.append(entry.family)
             }
+        }
+
+        if !failures.isEmpty {
+            print("Failed families: \(failures.joined(separator: ", "))")
+            exit(1)
         }
     }
 }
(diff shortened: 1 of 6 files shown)


Tests/CastTests/ChatTemplateTests.swift
```diff
@@ -0,0 +1,71 @@
...
+}
+
+private let chatTemplatePrompt = "Say hello as a JSON object with field 'text'."
+
+private func runGreeting(family: String, modelId: String) async throws {
+    do {
+        let model = try await CastModel.load(modelId)
+        let result: Greeting = try await model.cast(chatTemplatePrompt)
+        #expect(!result.text.isEmpty, "\(family) (\(modelId)) returned empty text field")
+        await model.unload()
```
jaylann: `CastModel.unload()` is synchronous (see `Sources/Cast/API/CastModel.swift:83` — `public func unload()`, no `async`). The `await` here will produce a Swift 6 warning (`no 'async' operations occur within 'await' expression`). Drop the `await` to match the pattern used in `Tests/CastTests/CastModelTests.swift:12` and `Examples/Sources/CallerManagedLoading/main.swift:34`.

Also: because `unload()` is only called on the success path (after `#expect`), a thrown error from `cast()` skips it. Not a true leak (ARC releases `model` when the function returns), but a `defer { model.unload() }` immediately after `let model = ...` makes the cleanup explicit and matches the surrounding code's intent.

Examples/Sources/ChatTemplates/main.swift
```diff
@@ -0,0 +1,50 @@
...
+
+        for entry in families {
+            print("--- \(entry.family) (\(entry.modelId))")
+            do {
+                let model = try await CastModel.load(entry.modelId)
+                let greeting: Greeting = try await model.cast(prompt)
+                print("\(entry.family): \(greeting)")
+                model.unload()
+            } catch {
+                print("\(entry.family) failed: \(error)")
```
jaylann, line 40: Catching the per-family error, printing it, and continuing means the example exits `0` even when one or more families fail. For an example whose only purpose is cross-family verification, this hides exactly the signal someone running it manually wants. Either:
- Track a `var hadFailure = false` and `exit(1)` (or `throw`) after the loop if any family failed, or
- Drop the per-iteration `do/catch` so the first failure surfaces with a non-zero exit.

Also: `model.unload()` is only reached on the success path. Move it to a `defer` at the top of the `do` so the model is released even when `cast()` throws — otherwise sequential iterations may hold an extra ~4 GB resident before ARC releases it across the catch boundary.

docs/decisions/0001-chat-template-handling.md
```diff
@@ -0,0 +1,71 @@
...
+- Gemma: `<start_of_turn>user ... <end_of_turn>`
+
+Each family ships a Jinja `chat_template` in its `tokenizer_config.json`.
+`MLXLMCommon`'s `UserInput` + `processor.prepare(input:)` already runs that
+template against the tokenizer when preparing model input.
+
+Cast's generation entry points (`Sources/Cast/API/CastModel+Generation.swift`,
+lines 118 and 197) compose a flat string of the form `"\(system)\n\n\(prompt)"`
+and hand it to `UserInput(prompt:)`. The downstream `processor.prepare(input:)`
+call then templates the result.
```
jaylann: Hard-coded line numbers (`lines 118 and 197`) will rot the first time `CastModel+Generation.swift` is touched (the lines are already approximate — `fullPrompt` is built around L117 and L197 today). Reference the symbol(s) instead, e.g. "the two `cast(...)` overloads in `CastModel+Generation.swift` build `fullPrompt` from `\"\\(system)\\n\\n\\(prompt)\"` before handing it to `UserInput(prompt:)`". The Consequences section's pointer at L70 ("the two call sites that build `fullPrompt`") is the right pattern — apply it here too.

jaylann (commented):
Overall: scope-correct verification PR. The critical CI-safety constraint is satisfied — all five new tests are `.requiresMetal`-gated, so CI won't try to load `default.metallib`. All five model ids resolve on Hugging Face (`HTTP 200` for each `mlx-community/...` repo). Vendored code (`Sources/MLXStructured/`) and the generation entry points (`CastModel+Generation.swift`) are untouched, matching the ADR's claim that this is a pure-verification change. ADR follows Status/Context/Decision/Consequences and the `Verification` subsection lists exactly the five families covered by the test file. CLAUDE.md learning is one line and references both the ADR path and test file path. Example's first line uses the required `// What this shows:` form and `Examples/Package.swift` registers the new target.

A handful of small but real issues below — none are blockers, but worth fixing before merge: a spurious `await` on the sync `unload()`, hard-coded line numbers in the ADR, and a verification example that silently swallows per-family failures. No `try!`/`as!`/force-unwraps in the changed files.

docs/decisions/0001-chat-template-handling.md
```diff
@@ -0,0 +1,71 @@
...
+  `tokenizer_config.json` works automatically — no Cast change required.
+- Family-specific quirks (whitespace normalization, missing `bos_token`, broken
+  Jinja templates, etc.) are `MLXLMCommon`'s responsibility. Cast surfaces such
+  failures as `CastError.generationFailed` rather than catching and rewriting
+  them.
+- Custom system prompts are joined with the user prompt by a single blank line.
+  Templates that draw a strong distinction between the system and user roles
+  (e.g. Llama's separate `system` header) still work because the templating
+  step sees a single user message and routes it correctly; we do not currently
+  expose a separate `system` channel through `UserInput`.
```
jaylann: "we do not currently expose a separate `system` channel through `UserInput`" — `MLXLMCommon.UserInput` does take structured `messages` as well as a flat `prompt`, so this is a Cast API choice rather than a downstream limitation. Worth one sentence noting that switching to the `messages:` initializer would be the future fix path if the blank-line join ever breaks a family — it keeps the Consequences section honest about why we picked the flat string.

Issue #39, opened 2026-03-20
CastBench: constrained vs unconstrained comparison mode


jaylann:
Compare constrained generation overhead against unconstrained.

```swift
let comparison = try await bench.compare(type: MovieReview.self, prompt: "Review Inception", iterations: 10)
print(comparison.overheadPercent)         // 2.8% slower
print(comparison.unconstrainedValidRate)  // 0.73 (73% valid without Cast)
```

PR #79, opened 2026-04-30
feat(tooling): CastBench benchmarking utility (constrained vs unconstrained) (#38, #39)

### LINKED ISSUE #38: CastBench: benchmarking utility (tok/s, latency, grammar overhead)
Built-in benchmarking utilities.

```swift
let bench = try await CastBench(model)
let result = try await bench.run(type: MovieReview.self, prompt: "Review Inception", iterations: 10)
print(result.tokensPerSecond)    // 45.2 tok/s
print(result.averageLatency)     // 1.23s
print(result.grammarOverheadMs)  // 2.1ms per token
print(result.averageTokenCount)  // 156 tokens
```

Output: formatted tables, Markdown, JSON.

jaylann:
## Summary
- New `CastBench` public type: `bench.run(...)` and `bench.compare(...)`.
- `BenchmarkResult` / `BenchmarkComparison` with table / Markdown / JSON output.
- Unconstrained path bypasses `GrammarMaskedLogitProcessor` for accurate overhead measurement.

Closes #38, #39.

## Test plan
- [x] swift build
- [x] swift test --filter CastBenchTests
- [ ] cd Examples && swift build (worktree-local SPM identifier limitation; will pass on CI which checks out into Cast/)
- [ ] Manual: swift run -c release CastBench (local only, requires model)

### DIFF
--- a/Sources/Cast/Bench/CastBench.swift
+++ b/Sources/Cast/Bench/CastBench.swift
@@ -2,7 +2,7 @@ import Foundation
 
 /// Output format for ``BenchmarkResult/formatted(as:)`` and
 /// ``BenchmarkComparison/formatted(as:)``.
-public enum OutputFormat: Sendable {
+public enum OutputFormat: String, Sendable, Codable {
     /// Fixed-width, human-readable plain-text table.
     case table
     /// GitHub-flavored Markdown table — paste into PR descriptions, READMEs, etc.
@@ -232,9 +232,10 @@ public struct CastBench: Sendable {
 
         var unconstrainedLatencies: [Duration] = []
         var unconstrainedTokens: [Int] = []
-        var validCount = 0
+        var unconstrainedSamples: [BenchmarkInstrumentation.IterationSample] = []
         unconstrainedLatencies.reserveCapacity(iterations)
         unconstrainedTokens.reserveCapacity(iterations)
+        unconstrainedSamples.reserveCapacity(iterations)
 
         for _ in 0 ..< iterations {
             let sample = try await BenchmarkInstrumentation.runUnconstrainedIteration(
@@ -245,9 +246,7 @@ public struct CastBench: Sendable {
             )
             unconstrainedLatencies.append(sample.latency)
             unconstrainedTokens.append(sample.tokenCount)
-            if sample.decoded != nil {
-                validCount += 1
-            }
+            unconstrainedSamples.append(sample)
         }
 
         let constrained = aggregate(
@@ -288,7 +287,7 @@ public struct CastBench: Sendable {
             constrained: constrainedWithOverhead,
             unconstrained: unconstrained,
             overheadPercent: overheadPct,
-            unconstrainedValidRate: Double(validCount) / Double(iterations)
+            unconstrainedValidRate: BenchmarkInstrumentation.validRate(of: unconstrainedSamples)
         )
     }
 
--- a/Sources/Cast/Bench/CastBench.swift
+++ b/Sources/Cast/Bench/CastBench.swift
@@ -133,7 +133,8 @@ public struct CastBench: Sendable {
     ///   - config: Sampling, timeout, and JSON-repair knobs forwarded to each
     ///     iteration's ``CastModel/cast(_:as:system:config:didGenerate:)-2yyul`` call.
     /// - Returns: Aggregated ``BenchmarkResult``.
-    /// - Throws: ``CastError`` from the underlying generation call. The first
+    /// - Throws: ``CastError/generationFailed(_:)`` when `iterations < 1`, or
+    ///   any ``CastError`` from the underlying generation call. The first
     ///   failing iteration aborts the run.
     public func run(
         type: (some Decodable & Sendable).Type,
@@ -141,7 +142,9 @@ public struct CastBench: Sendable {
         iterations: Int,
         config: CastConfiguration = CastConfiguration()
     ) async throws -> BenchmarkResult {
-        precondition(iterations >= 1, "CastBench.run requires iterations >= 1")
+        guard iterations >= 1 else {
+            throw CastError.generationFailed("CastBench.run requires iterations >= 1, got \(iterations)")
+        }
 
         var latencies: [Duration] = []
         var tokenCounts: [Int] = []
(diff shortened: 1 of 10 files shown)


Sources/Cast/Bench/CastBench.swift
```diff
@@ -0,0 +1,353 @@
...
+        for _ in 0 ..< iterations {
+            let sample = try await BenchmarkInstrumentation.runUnconstrainedIteration(
+                model: model,
+                type: type,
+                prompt: prompt,
+                config: config
+            )
+            unconstrainedLatencies.append(sample.latency)
+            unconstrainedTokens.append(sample.tokenCount)
+            if sample.decoded != nil {
```
jaylann: **Blocker — `unconstrainedValidRate` is always 1.0.**

`sample.decoded` is declared as `Bool` in `BenchmarkInstrumentation.IterationSample` (`Instrumentation.swift:18`), not `Bool?`. `decoded != nil` therefore always evaluates to `true`, so `validCount` always equals `iterations` and `unconstrainedValidRate = Double(validCount) / Double(iterations)` is hard-pinned to `1.0` regardless of whether the raw output actually parsed.

This defeats the entire point of `compare()`'s value proposition ("how much does the grammar buy you on this prompt?"). Should be:

```swift
if sample.decoded {
    validCount += 1
}
```

The Swift compiler should be emitting a warning here (`Comparing non-optional value of type 'Bool' to 'nil' always returns true`) — worth checking whether `-warnings-as-errors` is enabled in CI; if not, this slipped through silently.

The E2E assertion `comparison.unconstrainedValidRate <= 1.0` at `CastBenchE2ETests.swift:62` trivially passes against the constant `1.0`, so the test surface didn't catch this either. Please add a synthetic test that exercises the per-iteration validCount path (e.g. by feeding a known-bad output through `strictDecode` and asserting the resulting rate).

Sources/Cast/Bench/CastBench.swift
```diff
@@ -0,0 +1,353 @@
+import Foundation
+
+/// Output format for ``BenchmarkResult/formatted(as:)`` and
+/// ``BenchmarkComparison/formatted(as:)``.
+public enum OutputFormat: Sendable {
```
jaylann: Spec called for all four new public types (`CastBench`, `BenchmarkResult`, `BenchmarkComparison`, `OutputFormat`) to be both `Sendable` *and* `Codable`. `BenchmarkResult` and `BenchmarkComparison` conform; `OutputFormat` is `Sendable` only.

In practice nobody serializes the format selector, so this is minor — but if you want to keep the surface symmetric it's a one-liner:

```swift
public enum OutputFormat: String, Sendable, Codable {
    case table, markdown, json
}
```

String-backed makes the JSON representation stable (`"table"` rather than `{"table":{}}`).

Sources/Cast/Bench/Formatters.swift
```diff
@@ -0,0 +1,190 @@
...
+        var sectionRows = rows
+        sectionRows.append((
+            "Overhead",
+            String(format: "+%.2f%%", comparison.overheadPercent),
+            "—"
+        ))
+        sectionRows.append((
+            "Valid rate",
+            "100%",
+            String(format: "%.1f%%", comparison.unconstrainedValidRate * 100)
```
jaylann: Hard-coding `"100%"` for the constrained valid rate is fine *because* the grammar guarantees parseable output, but the assumption isn't stated anywhere and a future reader will reasonably wonder why the constrained side isn't computed. A one-line comment above this row (and the markdown equivalent at line 98) — e.g. `// Constrained valid rate is always 100% by construction (grammar masking guarantees a parseable parse).` — would prevent someone from "fixing" it later by wiring it to a real counter that doesn't exist.

Not blocking; just paying down the implicit-knowledge debt while it's fresh.

jaylann (commented):
## Review summary

Nice scaffolding overall — instrumentation correctly bypasses `GrammarMaskedLogitProcessor` by going through `MLXLMCommon.generate(...)` with a hand-built `TokenIterator`, the `.requiresMetal` gate is in the right place on `CastBenchE2ETests`, vendored `Sources/MLXStructured/` and `Sources/CMLXStructured/` are untouched, force-unwraps are absent, and the `TokenCounter` correctly tracks `max(seenCount)` rather than summing the cumulative-list sizes from `didGenerate`.

**Blocker:** `BenchmarkComparison.unconstrainedValidRate` is broken — see the inline on `CastBench.swift:248`. `sample.decoded` is `Bool`, so `decoded != nil` is always `true`, `validCount` always equals `iterations`, and the rate is hard-pinned to 1.0. That's the headline metric of `compare()` and it currently lies. The E2E test at `CastBenchE2ETests.swift:62` (`unconstrainedValidRate <= 1.0`) silently passes because everything passes — the bound check trivially holds against a constant.

Once that's fixed, the rest is minor: `OutputFormat` is missing `Codable` (the spec lists it alongside the other three public types), and the formatters hard-code `"100%"` for the constrained valid rate without commenting on why the assumption is safe. Both noted inline.

Request: please add at least one synthetic test that drives `compare()`'s validCount logic with a stub (e.g. inject decode results) so this regression class is caught without needing Metal. The current `CastBenchTests.swift` only round-trips already-constructed `BenchmarkComparison` values, which can't catch the bug.

No formatting nits, no scope creep — just the four bullets above. Not approving / not requesting changes per review protocol; flagging the blocker explicitly so it doesn't slip.

jaylann:
Addressed review feedback in 0d608f1:

- BLOCKER fixed: `BenchmarkComparison.unconstrainedValidRate` was hard-pinned to 1.0 via the `decoded \!= nil` tautology against a `Bool`. `compare()` now collects the unconstrained `IterationSample`s and delegates to a new `BenchmarkInstrumentation.validRate(of:)` helper that counts `sample.decoded == true`. The Swift warning (`Comparing non-optional value of type 'Bool' to 'nil' always returns true`) is gone.
- Synthetic regression coverage added in `Tests/CastTests/CastBenchTests.swift`: `validRateAllDecodedReturnsOne`, `validRateNoneDecodedReturnsZero`, `validRateMixedDecodedReturnsFraction`, `validRateEmptySamplesReturnsZero` — none require Metal, so they run on every CI build. Approach: extracted the rate computation to an internal `validRate(of:)` helper on `BenchmarkInstrumentation` so tests can drive it with synthetic samples without touching the public API.
- `OutputFormat` now `String, Sendable, Codable` for spec parity. The `String` raw representation keeps the JSON form stable (`"table"` rather than `{"table":{}}`).
- "100% by construction" assumption documented in both `formatTable` and `formatMarkdown` paths in `Sources/Cast/Bench/Formatters.swift`. Constant lifted to a named `constrainedValidRate` and rendered through the same `formatPercent` helper as the unconstrained side, so a future "fix" wiring it to a real counter would be visibly off the happy path.

`swift build` clean, `swift test --filter CastBenchTests` 16/16 pass (including the four new synthetic tests).

PR #80, opened 2026-04-30
feat(api): castStream() — AsyncSequence of PartialResult<T> (#35)

### LINKED ISSUE #35: Streaming generation: castStream() → AsyncSequence of PartialResult<T>
Streaming structured generation with partial results.

```swift
for try await partial in model.castStream("Analyze this", as: Analysis.self) {
    print(partial.title ?? "generating title...")
    print(partial.progress)  // 0.0 to 1.0
}
```

- `PartialResult<T>`: each field Optional, populated in JSON generation order
- `progress: Double` based on tokens generated vs expected
- Build on MLXStructured existing AsyncStream pattern

jaylann:
## Summary
- New `castStream(_:as:system:config:)` returning `AsyncThrowingStream<PartialResult<T>, Error>`.
- `@Castable` macro now synthesizes a nested `PartiallyGenerated` Optional-field mirror.
- Targeted addition to vendored `MLXStructured/Generate.swift` for a Grammar-input chunk stream (modification marker added per Apache-2.0 §4(b)).

Closes #35.

## Test plan
- [x] swift build
- [x] swift test --filter CastMacroTests (23 tests, including 6 new PartiallyGenerated expansion tests)
- [x] swift test --filter PartialResultDecodeTests (4 tests, progressive-fragment decode coverage)
- [x] swift test --filter CastTests (full library suite, 129 tests, all passing)
- [ ] cd Examples && swift build (worktree-local SPM identifier limitation; verified manually after merge / will pass on CI which checks out into Cast/)
- [ ] Manual: swift run -c release Streaming (local only, requires model)

### DIFF
--- a/Sources/Cast/API/CastModel+Stream.swift
+++ b/Sources/Cast/API/CastModel+Stream.swift
@@ -21,14 +21,22 @@ public extension CastModel {
     /// ```
     ///
     /// The stream terminates with ``CastError`` on schema/generation failure,
-    /// honors `Task.cancel()`, and respects ``CastConfiguration/timeout``.
+    /// honors `Task.cancel()` (surfaced as ``CastError/cancelled(partialOutput:)``
+    /// carrying the last decoded snapshot, if any), and respects
+    /// ``CastConfiguration/timeout``.
+    ///
+    /// Buffering uses ``AsyncThrowingStream/Continuation/BufferingPolicy/bufferingNewest(_:)``
+    /// with a small bound: a slow consumer paired with a fast model will see
+    /// the *latest* snapshots, not every intermediate one. Streaming UIs only
+    /// care about the most recent state, so this trades historical fidelity
+    /// for a memory ceiling.
     func castStream<T: Castable>(
         _ prompt: String,
         as type: T.Type = T.self,
         system: String? = nil,
         config: CastConfiguration = CastConfiguration()
     ) -> AsyncThrowingStream<PartialResult<T>, Error> {
-        AsyncThrowingStream { continuation in
+        AsyncThrowingStream(bufferingPolicy: .bufferingNewest(8)) { continuation in
             let task = Task {
                 do {
                     try await runStream(
@@ -88,6 +96,9 @@ public extension CastModel {
         let parameters = config.generateParameters
         let maxTokens = max(config.maxTokens, 1)
 
+        // Cross-isolation handle for the in-flight buffer so the cancellation
+        // catch (outside the perform closure) can recover the partial bytes.
+        let bufferHolder = StreamBufferHolder()
         do {
             try await withInFlightRegistration {
                 try await withGenerationTimeout(config.timeout) {
@@ -104,27 +115,38 @@ public extension CastModel {
 
                         var buffer = ""
                         var lastTokenCount = 0
+                        var lastProgress = 0.0
 
                         for try await chunk in stream {
                             if Task.isCancelled { break }
                             buffer.append(chunk.chunk)
                             lastTokenCount = chunk.totalTokens
+                            bufferHolder.set(buffer)
 
                             if let snapshot = decodePartial(
                                 T.self,
                                 from: buffer,
                                 tokenCount: lastTokenCount,
-                                maxTokens: maxTokens
+                                maxTokens: maxTokens,
+                                minProgress: lastProgress
                             ) {
+                                lastProgress = snapshot.progress
                                 continuation.yield(snapshot)
                             }
                         }
+
+                        // Cancellation must not be reported as a decode/repair
+                        // failure: a half-formed buffer almost always fails
+                        // `yieldTerminal`, masking the real cause. Surface the
+                        // outer `Task.isCancelled` check below instead.
+                        if Task.isCancelled { return }
 
                         try yieldTerminal(
                             T.self,
                             buffer: buffer,
                             tokenCount: lastTokenCount,
                             maxTokens: maxTokens,
+                            minProgress: lastProgress,
                             config: config,
                             continuation: continuation
                         )
@@ -136,7 +158,7 @@ public extension CastModel {
             throw error
         } catch is CancellationError {
             cleanupGPU()
-            throw CastError.cancelled(partialOutput: nil)
+            throw CastError.cancelled(partialOutput: bufferHolder.snapshot())
         } catch {
             cleanupGPU()
             throw CastError.generationFailed(error.localizedDescription)
@@ -149,8 +171,29 @@ public extension CastModel {
         }
 
         if Task.isCancelled {
-            throw CastError.cancelled(partialOutput: nil)
+            throw CastError.cancelled(partialOutput: bufferHolder.snapshot())
         }
+    }
+}
+
+/// Sendable bridge for the in-flight stream buffer. Lets the post-cancel
+/// `catch` block recover the last bytes accumulated inside the perform
+/// closure (which runs on a different actor) so they can ride along on
+/// `CastError.cancelled(partialOutput:)`.
+private final class StreamBufferHolder: @unchecked Sendable {
+    private let lock = NSLock()
+    private var buffer = ""
+
+    func set(_ newValue: String) {
+        lock.lock()
+        defer { lock.unlock() }
+        buffer = newValue
+    }
+
+    func snapshot() -> String? {
+        lock.lock()
+        defer { lock.unlock() }
+        return buffer.isEmpty ? nil : buffer
     }
 }
 
@@ -158,7 +201,8 @@ private func decodePartial<T: Castable>(
     _: T.Type,
     from buffer: String,
     tokenCount: Int,
-    maxTokens: Int
+    maxTokens: Int,
+    minProgress: Double
 ) -> PartialResult<T>? {
     let candidate: String
     switch JSONRepair.repair(buffer) {
@@ -177,7 +221,8 @@ private func decodePartial<T: Castable>(
         return nil
     }
 
-    let progress = min(Double(tokenCount) / Double(maxTokens), 1.0)
+    let raw = min(Double(tokenCount) / Double(maxTokens), 1.0)
+    let progress = max(raw, minProgress)
     return PartialResult<T>(value: value, progress: progress, tokenCount: tokenCount)
 }
 
@@ -186,6 +231,7 @@ private func yieldTerminal<T: Castable>(
     buffer: String,
     tokenCount: Int,
     maxTokens: Int,
+    minProgress: Double,
     config: CastConfiguration,
     continuation: AsyncThrowingStream<PartialResult<T>, Error>.Continuation
 ) throws {
@@ -216,6 +262,7 @@ private func yieldTerminal<T: Castable>(
         )
     }
 
-    let progress = min(Double(tokenCount) / Double(maxTokens), 1.0)
+    let raw = min(Double(tokenCount) / Double(maxTokens), 1.0)
+    let progress = max(raw, minProgress)
     continuation.yield(PartialResult<T>(value: value, progress: progress, tokenCount: tokenCount))
 }
--- a/Sources/Cast/API/CastModel+Stream.swift
+++ b/Sources/Cast/API/CastModel+Stream.swift
@@ -26,17 +26,22 @@ public extension CastModel {
     /// ``CastConfiguration/timeout``.
     ///
     /// Buffering uses ``AsyncThrowingStream/Continuation/BufferingPolicy/bufferingNewest(_:)``
-    /// with a small bound: a slow consumer paired with a fast model will see
-    /// the *latest* snapshots, not every intermediate one. Streaming UIs only
-    /// care about the most recent state, so this trades historical fidelity
-    /// for a memory ceiling.
+    /// with bound `1`: a slow consumer paired with a fast model will see only
+    /// the *most recent* snapshot at any time, not every intermediate one.
+    /// Streaming UIs only care about the latest state, so this trades historical
+    /// fidelity for a fixed memory ceiling. The terminal yield is the final
+    /// write before `.finish()` and is therefore never dropped.
     func castStream<T: Castable>(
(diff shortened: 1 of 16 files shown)


jaylann (commented):
Streaming PR review.

**No blocker-class issues found** in the highest-priority checks:
- Vendored `Sources/MLXStructured/Generate.swift` preserves the upstream `Created by Ivan Petrukha` header AND has the required `// Modifications: add Grammar-input chunk stream by Justin Lanfermann, 2026` line.
- `@attached(member, names:)` correctly lists `named(PartiallyGenerated)` so the synthesized type is visible to consumers.
- `castStream` is wrapped in `withInFlightRegistration` + `withGenerationTimeout` and registers `continuation.onTermination` to cancel the inner Task — cancellation envelope mirrors `cast()`.
- `Tests/CastTests/CastStreamTests.swift` is gated with `.requiresMetal`.
- `Package.swift` test deps for `CastTests` now include `JSONSchema` + `Collections`.
- Submodule pointer (`Sources/CMLXStructured/xgrammar`) is unchanged vs `stage`.
- `PartialResult<T>` is `Sendable`; `Castable` requires `PartiallyGenerated: Sendable & Decodable` so the constraint flows.
- DocC topic added (`<doc:Streaming>`), example registered in `Examples/Package.swift`, first-line `// What this shows:` comment present.
- Macro expansion suite covers simple, wrapper-decorated, nested `@Castable`, primitive-array, and nested-array element cases.

Actionable findings are inline. The most important is the cancellation-during-stream codepath: `break` exits the for-loop but then unconditionally calls `yieldTerminal` against a half-decoded buffer, which can throw `repairFailed`/`decodingFailed` instead of `CastError.cancelled`. The other findings are smaller (token-count drift never reaches consumers; nullable/optional-field handling for partials; one DocC cross-link gap).

No formatting nits.

Sources/Cast/API/CastModel+Stream.swift
```diff
@@ -0,0 +1,221 @@
...
+                        }
+
+                        try yieldTerminal(
+                            T.self,
+                            buffer: buffer,
+                            tokenCount: lastTokenCount,
+                            maxTokens: maxTokens,
+                            config: config,
+                            continuation: continuation
+                        )
```
jaylann, line 157: **Cancellation path silently mis-reports as a decode/repair error.** When the consumer cancels mid-stream, `if Task.isCancelled { break }` (line 109) exits the for-loop, then control falls through to `yieldTerminal(...)` which decodes the *partial* buffer. On a half-formed buffer that almost always throws `CastError.repairFailed` or `CastError.decodingFailed`, which is then caught by `catch let error as CastError` at line 134 and rethrown — so the consumer sees a misleading error instead of `CastError.cancelled`. The post-loop `if Task.isCancelled { throw CastError.cancelled(...) }` at line 151 is unreachable on this path.

Fix: gate the terminal call on cancellation, e.g.

```swift
if Task.isCancelled {
    throw CancellationError()
}
try yieldTerminal(...)
```

or return early from the closure and let the outer `if Task.isCancelled` at line 151 fire.

Sources/Cast/API/CastModel+Stream.swift
```diff
@@ -0,0 +1,221 @@
...
+
+    guard let value = try? JSONDecoder().decode(
+        T.PartiallyGenerated.self,
+        from: Data(candidate.utf8)
+    ) else {
+        return nil
+    }
+
+    let progress = min(Double(tokenCount) / Double(maxTokens), 1.0)
+    return PartialResult<T>(value: value, progress: progress, tokenCount: tokenCount)
```
jaylann, line 231: **`tokenCount` may be 0 for the first yields when chunks are byte-not-token granular.** `lastTokenCount` is initialized to 0 and only updated from `chunk.totalTokens`, which is itself a chunk-counter increment in `MLXStructured.generateStream`. If MLXLMCommon ever yields a `.chunk("")` or coalesces multiple tokens into one chunk, `progress` will be misleading. Not a correctness bug today but worth considering: clamp `progress` to `max(progress, previousProgress)` so the public API guarantee of monotonic progress (which `CastStreamTests` asserts) holds across runtime changes upstream.

Sources/MLXStructured/Generate.swift
```diff
@@ -24,79 +25,143 @@ public func generate(
...
+                switch generation {
+                case let .chunk(text):
+                    // Approximate token count by chunk count — `Generation.chunk`
+                    // is one decoded token's text per yield in MLXLMCommon's
+                    // current pipeline. Exact counts come via `.info` at the
+                    // tail; downstream consumers use this only for `progress`.
+                    totalTokens += 1
+                    continuation.yield(GrammarChunk(chunk: text, totalTokens: totalTokens))
+                case let .info(info):
+                    totalTokens = info.generationTokenCount
```
jaylann, line 90: **`.info` updates `totalTokens` but never yields a corrected chunk.** When `case let .info(info)` arrives, `totalTokens = info.generationTokenCount` overrides the chunk-counted value — but no `GrammarChunk` is yielded after. The corrected count therefore never reaches downstream consumers; the last yielded `GrammarChunk` carries the (potentially stale) chunk-counter total.

If the intent is for `.info` to be an authoritative correction, yield a final empty-text chunk: `continuation.yield(GrammarChunk(chunk: "", totalTokens: info.generationTokenCount))`. Otherwise drop the `totalTokens = info.generationTokenCount` line — it's currently dead code.

Sources/MLXStructured/Generate.swift
```diff
@@ -24,79 +25,143 @@ public func generate(
...
+            var totalTokens = 0
+            for await generation in upstream {
+                if Task.isCancelled { break }
+                switch generation {
+                case let .chunk(text):
+                    // Approximate token count by chunk count — `Generation.chunk`
+                    // is one decoded token's text per yield in MLXLMCommon's
+                    // current pipeline. Exact counts come via `.info` at the
+                    // tail; downstream consumers use this only for `progress`.
+                    totalTokens += 1
```
jaylann, line 84: Worth documenting at the public API level (`generateStream`'s docstring) that `totalTokens` is approximate (chunk-counted, not real token-counted). Consumers reading just the doc comment on `GrammarChunk` see `public let totalTokens: Int` with no caveat — the approximation comment lives only inside the implementation. A one-liner on `GrammarChunk.totalTokens` ("approximate, derived from chunk count") would prevent misuse.

Sources/CastMacros/CastableMacro.swift
```diff
@@ -280,6 +281,55 @@ private func generateInitDecl(properties _: [PropertyInfo]) -> DeclSyntax {
...
+/// `Castable` and routed through `.PartiallyGenerated`. Arrays carry the
+/// projection through to their element type. The whole field is then made
+/// Optional regardless of whether it was Optional originally — partial
+/// snapshots have no notion of "required".
+private func partialFieldType(for prop: PropertyInfo) -> String {
+    if prop.isArray {
+        let element = prop.arrayElementType ?? "String"
+        return "[\(partialElementType(element))]?"
+    }
+    return "\(partialBaseType(prop.typeName))?"
```
jaylann, line 350: **Optional fields in the source struct collapse to non-Optional in the partial.** `partialFieldType` always appends `?`, so a source field declared `var nickname: String?` becomes `var nickname: String?` in `PartiallyGenerated` — which is fine for the type, but `partialBaseType(prop.typeName)` is called with the *unwrapped* type name (because `parseType` strips the Optional). For a field `var nested: Inner?` where `Inner` is `@Castable`, you correctly get `Inner.PartiallyGenerated?`. Good. **However** the macro expansion test suite has no case for an Optional source field — please add one to lock this in (e.g. `var subtitle: String?` → `var subtitle: String?`, and `var nested: Inner?` → `var nested: Inner.PartiallyGenerated?`).

Sources/CastMacros/CastableMacro.swift
```diff
@@ -280,6 +281,55 @@ private func generateInitDecl(properties _: [PropertyInfo]) -> DeclSyntax {
...
+}
+
+private func partialBaseType(_ typeName: String) -> String {
+    if stringTypes.contains(typeName) || numericTypes.contains(typeName) || boolTypes.contains(typeName) {
+        return typeName
+    }
+    return "\(typeName).PartiallyGenerated"
+}
```
jaylann, line 359: **Nested `@Castable` resolution is a textual `"\(typeName).PartiallyGenerated"`** — so a property typed by a fully-qualified name (e.g. `var x: SomeModule.Inner`) or a typealias to a `Castable` will produce `SomeModule.Inner.PartiallyGenerated` (probably fine) or `MyAlias.PartiallyGenerated` (works only if the alias is to a `Castable`). And any field whose type is *not* a primitive and *not* `@Castable` (e.g. `var date: Date`, `var url: URL`, an enum) compiles to `Date.PartiallyGenerated` which doesn't exist → consumer gets an opaque error.

At minimum, add a diagnostic in `validateProperties` for fields whose type isn't in the known primitive set and isn't structurally `@Castable`-able (best-effort: warn that the field will be projected as `.PartiallyGenerated` and require manual conformance). Or document the limitation prominently on the `@Castable` macro doc.

Sources/Cast/API/PartialResult.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+
+/// A snapshot of a still-streaming ``CastModel/castStream(_:as:system:config:)`` call.
+///
+/// `value` is the partially-decoded form of `T` (every field Optional, fields
+/// that haven't been emitted yet are `nil`). `progress` reflects how close the
+/// generation is to its `maxTokens` budget, clamped to `0...1`. `tokenCount`
+/// is the number of tokens produced so far.
+public struct PartialResult<T: Castable>: Sendable {
```
jaylann, line 9: Cross-link missing: the `PartialResult` doc comment references `CastModel/castStream(_:as:system:config:)` but the existing `cast()` doc comment in `CastModel+Generation.swift` does not mention `castStream` as a streaming alternative. Add a `- SeeAlso: ``CastModel/castStream(_:as:system:config:)``` in the `cast()` docstring (or a short paragraph) so DocC surfaces the relationship in both directions.

Sources/Cast/API/CastModel+Stream.swift
```diff
@@ -0,0 +1,221 @@
...
+                    continuation.finish()
+                } catch {
+                    continuation.finish(throwing: error)
+                }
+            }
+            continuation.onTermination = { _ in
+                task.cancel()
+            }
+        }
+    }
```
jaylann, line 63: **`AsyncThrowingStream` constructed without explicit buffering policy uses `.unbounded`.** A slow consumer of `castStream` against a fast model on M-series hardware will accumulate every per-token snapshot in memory until consumption. For long generations (`maxTokens` ~ 4k) plus large `T.PartiallyGenerated` payloads this is a memory hazard.

Consider `AsyncThrowingStream(bufferingPolicy: .bufferingNewest(1))` so back-pressure drops stale snapshots — for a streaming UI use case, only the latest snapshot matters. The terminal yield will not be dropped because it's the last write before `.finish()`.

jaylann:
Addressed remaining review feedback:

- **Buffering policy** (`Sources/Cast/API/CastModel+Stream.swift`): switched from `.bufferingNewest(8)` to `.bufferingNewest(1)`. A slow consumer paired with a fast generation now sees only the most recent snapshot rather than buffering up to 8. Updated the doc-comment accordingly. Terminal yield is the last write before `.finish()` so it can never be dropped — `CastStreamTests.castStreamProducesPartialsAndTerminal` still holds (it asserts `>= 2` snapshots, which a tight `for try await` loop satisfies even under bound-1 back-pressure since the terminal write is preserved).
- **CastableMacro warning** (`Sources/CastMacros/CastableDiagnostic.swift`, `Sources/CastMacros/CastableMacro.swift`): added `unknownNonPrimitiveType(typeName:)` warning emitted when a field is a known-suspect Foundation value type (`Date`, `URL`, `UUID`, `Data`, `Decimal`, `TimeInterval`). The diagnostic explains the field will be projected as `<TypeName>.PartiallyGenerated?` and how to fix it (wrap in a `@Castable` struct or pre-convert to a primitive). I deliberately did not warn on every non-primitive — that would flood the build for legitimate nested `@Castable` structs the macro can't see. Restructured `CastableDiagnostic` to per-case severity so warnings no longer block expansion.
- **Macro tests** (`Tests/CastMacroTests/CastDiagnosticTests.swift`): added `unknownFoundationType` covering `var when: Date`, asserting both that the warning fires and that expansion still succeeds with the synthesized `Date.PartiallyGenerated?` mirror.

Items already addressed in the prior commit (`afdbe07`) on this branch and verified intact:
- Cancellation gate before `yieldTerminal` (returns from the perform closure so the post-`catch` `Task.isCancelled` check fires `CastError.cancelled(partialOutput:)` instead of misleading decode errors).
- `Generate.swift` `.info` arm yields a final `GrammarChunk(chunk: "", totalTokens: info.generationTokenCount)` so the corrected count reaches consumers; doc-comment on `GrammarChunk.totalTokens` notes the chunk-counted approximation. Vendored attribution headers preserved (`Created by Ivan Petrukha` + `Modifications:` line untouched).
- `cast()` doc-comment cross-links `castStream()` with `- SeeAlso:`.

Verification:
- `swift build` clean (1 pre-existing `Grammar` Sendable warning, unrelated)
- `swift test --filter CastMacroTests`: 25/25 passed
- Full `swift test` (CI=true): 168/168 passed
- `Examples/` build skipped locally — worktree dir name (`streaming`) breaks SPM's `.package(path: "..")` identifier resolution as documented in `CLAUDE.md`; CI builds Examples in a `Cast/` checkout where this works.

jaylann:
Note on the `tokenCount` monotonic-clamp suggestion: not applied — current `max(raw, minProgress)` floor + the monotonic-progress assertion in `CastStreamTests` are the safety net today. Will revisit if MLXLMCommon ever starts emitting zero-token chunks.

Rebased on top of merged #78. CI re-running.

Issue #90, opened 2026-05-01
fix(api): annotations() failure silently degrades to empty dict in 5 hot paths


jaylann:
Five call sites use `(try? SchemaGenerator.annotations(for: type)) ?? [:]`, which means **any** failure to extract field annotations results in a prompt that silently lacks descriptions/examples — degrading output quality with no signal to the caller:

- \`Sources/Cast/API/CastModel+Generation.swift:44\`
- \`Sources/Cast/API/CastModel+Generation.swift:80\`
- \`Sources/Cast/API/CastModel+Extract.swift:44\`
- \`Sources/Cast/API/CastModel+Stream.swift:85\`
- \`Sources/Cast/Bench/Instrumentation.swift:84\`

This is the same anti-pattern as #85 (PromptEngine's silent \`[Schema encoding failed]\` fallback) — both hide real failures behind a happy-looking empty default.

## Suggested fix

Bubble the error. \`cast()\`, \`castStream()\`, \`extract()\` already throw \`CastError\`, so let \`SchemaGenerator.annotations(for:)\`'s error surface to the caller (or wrap as \`CastError.schemaGenerationFailed\`):

```swift
let annotations = try SchemaGenerator.annotations(for: type)
```

For \`Bench/Instrumentation.swift\`, decide whether bench failures should also throw or whether logging + benchmark-marker is acceptable; document the choice.

## Verification

- After the fix, deliberately breaking annotations (e.g. wrapping a non-Castable type) produces a thrown \`CastError\` instead of a silently-degraded prompt.
- \`rg -nF 'try? SchemaGenerator.annotations' Sources/\` returns zero hits.

## Related

- #85 — PromptEngine's parallel silent fallback at \`PromptEngine.swift:31, 119\`.

Issue #89, opened 2026-05-01
fix(schema): replace force-cast in SchemaGenerator (as\! any Sendable)


jaylann:
`Sources/Cast/Schema/SchemaGenerator.swift:88` uses a force-cast:

```swift
mirrorInstance = castableType.init() as\! any Sendable
```

This is the only `as\!` in `Sources/Cast` or `Sources/CastMacros`. `Castable` already requires `Sendable` (`Sources/Cast/API/Castable.swift:26`), so the cast is *expected* to succeed — but it traps the process if it ever doesn't, instead of surfacing a clean `CastError.schemaGenerationFailed`.

## Suggested fix

Two options:

1. Tighten the protocol: change the synthesized `init()` to be statically typed as returning `any Sendable` (or constrain `castableType` so the compiler accepts the cast without `\!`).
2. Convert to a guarded `as?` that throws `CastError.schemaGenerationFailed("…not Sendable…")`. Cheaper to land and consistent with the rest of `SchemaGenerator`'s error handling.

Prefer (1) if it's tractable from the macro side; otherwise (2) is acceptable.

## Verification

- `rg -n 'as\!|try\!|fatalError\(' Sources/Cast Sources/CastMacros` reports zero hits in production (currently 1).
- Existing `SchemaGeneratorTests.swift` tests still pass.

Issue #88, opened 2026-05-01
ci: fail PR when committed Cast.docc/Examples/* drifts from generator output


jaylann:
`scripts/generate-example-docs.sh` is run automatically by `.github/workflows/docs.yml` during DocC deploy from `main`, but never enforced as a pre-merge gate. As a result, committed articles in `Sources/Cast/Cast.docc/Examples/` can silently drift from the source-of-truth `Examples/Sources/<Name>/main.swift` files. (Concrete example: `ChatTemplates.md` is missing on `stage` — tracked in #87.)

## Suggested fix

Add a step to `.github/workflows/test.yml` (or a new `docs-drift.yml`) that fails when the script's output diverges from what's checked in:

```yaml
- name: Check DocC examples are in sync
  run: |
    ./scripts/generate-example-docs.sh
    git diff --exit-code Sources/Cast/Cast.docc/Examples/
```

This runs on every PR targeting `stage`, so contributors notice the drift before merge.

## Verification

- After landing this PR plus the ChatTemplates regeneration, `git diff --exit-code Sources/Cast/Cast.docc/Examples/` is clean.
- A deliberate edit to `Examples/Sources/HelloCast/main.swift` without regenerating triggers the new gate locally.

Issue #87, opened 2026-05-01
docs(docc): add missing ChatTemplates.md article (broken <doc:ChatTemplates> link on stage)


jaylann:
`Sources/Cast/Cast.docc/Cast.md` lists `<doc:ChatTemplates>` under `## Topics > Examples`, but `Sources/Cast/Cast.docc/Examples/ChatTemplates.md` does **not** exist on `stage`. The corresponding example source `Examples/Sources/ChatTemplates/main.swift` is present.

## Why it matters

- Local `swift package generate-documentation` builds with broken cross-reference.
- DocC published from `main` is fine because `.github/workflows/docs.yml` runs `scripts/generate-example-docs.sh` before building, but the **stage** branch DocC preview is broken.
- Sets up future regressions if the generator script's `find -delete` is ever skipped.

## Suggested fix

```bash
./scripts/generate-example-docs.sh
git add Sources/Cast/Cast.docc/Examples/ChatTemplates.md
git commit -m "docs(docc): regenerate ChatTemplates article"
```

Pair this with #-CI-DRIFT-GATE (separate issue) so the article cannot drift again.

## Verification

- `diff <(ls Sources/Cast/Cast.docc/Examples/) <(ls Examples/Sources/)` → empty diff after fix.
- `swift package generate-documentation --target Cast` builds without "missing reference" warnings on `<doc:ChatTemplates>`.

PR #102, opened 2026-05-01
fix(stage): bundle stale-state cleanup (#86, #87, #88, #89, #90)

### LINKED ISSUE #86: docs(readme): refresh roadmap — streaming, bench, chat templates already shipped
The `README.md` narrates a state of the project that predates the current `stage` HEAD. Several features it labels "in flight" are already shipped and tested.

## Stale claims

- **Lines ~10** ("Status: pre-1.0. Phases 0–2 ... shipped. Phase 3 in progress: ... streaming, benchmarks, per-model chat templates ... still in flight"):
  - `castStream` is implemented in `Sources/Cast/API/CastModel+Stream.swift` (288 LoC, tests in `Tests/CastTests/CastStreamTests.swift`). Issue #35 is **closed**.
  - `CastBench` is implemented in `Sources/Cast/Bench/{CastBench,Formatters,Instrumentation}.swift`. Issues #38 and #39 are **closed**.
  - Per-model chat templates verified across 5 families in `Tests/CastTests/ChatTemplateTests.swift`. Issue #37 is **closed**.
  - "Phases 0–2" framing references retired phase labels (taxonomy migrated 2026-04-30; see `.claude/rules/pr-conventions.md`).
- **Line ~64** ("Per-model chat templates are tracked in #37"): #37 is closed.
- **Lines ~73–75** ("Coming with #38 ... and #39"): both closed.

## Suggested fix

- Drop the "Phase N" framing. Describe the current public surface flatly: `cast`, `castStream`, `classify`, `extract`, `prepare`, lifecycle, GPU safety, timeout/cancellation, JSON repair, CastBench, 5-family chat templates, DocC site.
- Replace the closed-issue links with #44 (iOS demo app) for the genuine "in flight" content.
- Update the "Benchmarks" section to point at `Sources/Cast/Bench/CastBench.swift` and the DocC `<doc:CastBench>` article instead of #38/#39.

## Verification

- `gh issue view 35 -q .state` / 37 / 38 / 39 → all `CLOSED`.
- `git grep '#35\|#37\|#38\|#39' README.md` returns nothing after the fix.

jaylann:
## Summary

Bundles five `priority:high` stage-cleanup issues plus a light README polish into one branch.

- **#89** `fix(schema)`: drops the only `as\!` in production source (`SchemaGenerator.swift:88`). `mirrorInstance` is now typed `Any` (which is what `Mirror(reflecting:)` takes), so no cast is needed at all.
- **#90** `fix(api)`: replaces `(try? SchemaGenerator.annotations(for: type)) ?? [:]` with a real `try` that surfaces failures as `CastError.schemaGenerationFailed`. Five sites: `CastModel+Generation.swift` (×2), `CastModel+Extract.swift`, `CastModel+Stream.swift`, `Bench/Instrumentation.swift`. No more silently-degraded prompts.
- **#87** `docs(docc)`: regenerated all `Cast.docc/Examples/*.md` via the existing `scripts/generate-example-docs.sh`. Adds the missing `ChatTemplates.md` (the `<doc:ChatTemplates>` link was broken on stage) and pulls `Extract.md` / `Streaming.md` back in sync with `Examples/Sources/`.
- **#88** `ci`: adds a drift-check step to `.github/workflows/test.yml` that runs the generator and `git diff --exit-code`s the output. Future drift fails PR CI before merge.
- **#86** `docs(readme)`: drops closed-issue references (#35/#37/#38/#39), retires the "Phase 0–2 / Phase 3" framing, surfaces `castStream` / `extract` in the generation-modes block, points the Benchmarks section at `CastBench` + DocC, and adds shields.io badges + a one-line TOC.

## Test plan

- [x] `swift build` clean
- [x] `swift test` — 200 tests pass (CI=true)
- [x] `rg -n 'as\\!|try\\!|fatalError\(' Sources/Cast Sources/CastMacros` → 0 hits (#89)
- [x] `rg -nF 'try? SchemaGenerator.annotations' Sources/` → 0 hits (#90)
- [x] `diff <(ls Cast.docc/Examples/) <(ls Examples/Sources/)` → match (#87)
- [x] `rg -n '#35|#37|#38|#39' README.md` → 0 hits (#86)
- [ ] CI: new drift gate runs on this PR and is green
- [ ] CI: `pr-conventions.yml` accepts label set

Closes #86, #87, #88, #89, #90.


.github/workflows/test.yml
```diff
@@ -32,3 +32,11 @@ jobs:
 
       - name: Run Tests
         run: swift test
+
+      - name: Verify DocC examples are in sync with Examples/Sources/
+        run: |
+          ./scripts/generate-example-docs.sh
+          if ! git diff --exit-code Sources/Cast/Cast.docc/Examples/; then
```
jaylann, line 39: `git diff --exit-code` only flags **modifications and deletions of tracked files** — it ignores untracked files. The generator script does `find -delete` before regenerating, so existing committed articles round-trip fine and stale committed articles surface as deletions. But if a contributor adds a new `Examples/Sources/<New>/main.swift` and forgets to commit the generated `Cast.docc/Examples/<New>.md`, the regenerated `<New>.md` is **untracked** and this gate passes silently. Use `git status --porcelain Sources/Cast/Cast.docc/Examples/` (or `git ls-files --others --exclude-standard` plus the diff check) to also catch added files.

README.md
```diff
@@ -85,14 +94,25 @@ with Cast's grammar-constrained decoding. The first one is a good default.
...
+for measuring tok/s, latency, grammar-masking overhead, and (optionally)
+constrained-vs-unconstrained validity rates on your own prompts and types.
+
+```swift
+let bench = CastBench(model)
+let result = try await bench.run(type: Person.self, prompt: "...", iterations: 5)
+print(result.formatted(as: .markdown))
+```
+
+See `Sources/Cast/Bench/CastBench.swift` and the [DocC article](https://jaylann.github.io/Cast/documentation/cast/castbench-article)
```
jaylann: The `castbench-article` slug doesn't exist on the published DocC site (`curl -I https://jaylann.github.io/Cast/documentation/cast/castbench-article` → 404). DocC routes `Examples/CastBench.md` to the article path; with a same-named symbol `CastBench` (`Sources/Cast/Bench/CastBench.swift`), the lowercased slug collides. Verify the actual article URL on the published site (likely `/cast/castbench` once it republishes — but note the symbol page is *also* at that slug, so DocC may have an actual collision worth resolving by renaming the article).

Sources/Cast/Bench/Instrumentation.swift
```diff
@@ -74,14 +74,15 @@ enum BenchmarkInstrumentation {
         // overhead numbers compare apples to apples. Surface schema-generation
         // failures as `CastError` for parity with `CastModel.cast(...)`.
         let schema: JSONSchema
+        let annotations: [String: FieldAnnotation]
         do {
             schema = try SchemaGenerator.schema(for: type)
+            annotations = try SchemaGenerator.annotations(for: type)
```
jaylann, line 80: Minor: `SchemaGenerator.schema(for:)` and `annotations(for:)` both delegate to the private `cached(_:)` which only ever throws whatever `build(_:)` / `ZeroSchemaDecoder.decode` throws — none of those are `CastError` (`rg 'CastError\.' Sources/Cast/Schema` returns 0 hits). The `catch let error as CastError { throw error }` branch above (line 81) is therefore dead in this codebase. Not introduced by this PR, just observing it's still here. Either drop it or wait until `SchemaGenerator` actually starts vending typed errors.

jaylann:
## Review summary

Three inline comments above. Quick rundown of what I checked and didn't find issues with, plus a couple of general notes that don't fit on a specific line.

**Looks correct:**
- **#90 refactor**: All five sites correctly fold `schema` + `annotations` into one `do/catch`. `SchemaGenerator.cached(_:)` is the shared cache, so the second call is a lock-protected dict lookup — no perf hit. Behavior on the happy path is unchanged. `schemaGenerationFailed` now correctly fires when annotation extraction fails (it didn't before — `try?` swallowed it).
- **#89 fix**: `mirrorInstance: Any` is fine. The local is consumed only by `Mirror(reflecting:)` (which takes `Any`) and the `for child in mirror.children` loop, neither of which depend on `Sendable`. `build(_:)` runs synchronously inside `cached(_:)`'s lock — no Sendable hop. `info.zeroInstance` is `any Sendable` and `Castable.init()` returns a `Sendable` value, so we're not actually losing safety; we're just dropping a redundant existential cast.
- **#86 README**: `CONTRIBUTING.md` and `MIGRATION.md` both exist at repo root, so the new references resolve.

**General notes (not blocking):**

1. **No regression test for #89.** The original plan mentioned extending `Tests/CastTests/SchemaGeneratorTests.swift` but no test landed. Existing tests already exercise both branches (`SimpleStruct` for non-Castable, `ConstrainedStruct` / `EnumFieldStruct` for Castable) and would have caught a behavioral regression — but a one-line test asserting that schema generation for a `Castable` type with a wrapper still preserves the wrapper's `maxLength` would lock in the fix's intent and prevent silent regressions. Worth a follow-up.

2. **PR labels.** Single `type:bug` is undersold given the PR also lands a new CI gate (#88), README rewrite (#86), and a brand-new DocC article (`ChatTemplates.md`, #87). Per `.claude/rules/pr-conventions.md` exactly one `type:*` is allowed, so this isn't a CI gate failure — but `type:enhancement` or `type:chore` would more accurately reflect the bundle. `type:bug` will land in release notes under "Bug fixes" and miss the user-facing README/DocC/CI improvements. Not a blocker; flagging for next time.

3. **Generated article copy.** `ChatTemplates.md` and `Streaming.md` start with a lowercased word ("the same prompt…", "progressive partial-result yields…") because `scripts/generate-example-docs.sh`'s `sed '1s/^What this shows: //'` strips the prefix without re-capitalizing. Cosmetic; pre-existing pattern; don't fix in this PR.

.github/workflows/test.yml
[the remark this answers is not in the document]
jaylann: Fixed in 8b3909f: added a `git status --porcelain Sources/Cast/Cast.docc/Examples/` untracked-file check immediately after the existing `git diff --exit-code`. The diff check still catches modifications/deletions of tracked articles; the new check catches the case where a contributor adds `Examples/Sources/<New>/main.swift` and forgets to commit the regenerated `Cast.docc/Examples/<New>.md`.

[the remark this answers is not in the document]
jaylann: Fixed in 8b3909f: added a `git status --porcelain Sources/Cast/Cast.docc/Examples/` untracked-file check after the existing `git diff --exit-code`. Untracked regenerated articles now fail the gate.

jaylann: Fixed in 8b3909f: added a `git status --porcelain Sources/Cast/Cast.docc/Examples/` untracked-file check after the existing `git diff --exit-code`. Same root cause as the P2 / Copilot threads above — handled together.

README.md
jaylann: Fixed in 8b3909f: confirmed both candidate URLs return HTTP/2 404 against the deployed DocC site (`/documentation/cast/castbench` and `/documentation/cast/castbench-article`), and `castbench` does not appear in `https://jaylann.github.io/Cast/index/index.json` at all — the published site lags `stage` and CastBench docs have not shipped yet. Combined with the slug-collision learning (article + symbol both want `castbench`), hard-coded deep links would remain unstable even after the next release. Dropped both deep DocC links and now point at the source file plus the Cast DocC root with breadcrumb to *Examples → CastBench*.

Sources/Cast/Bench/Instrumentation.swift
jaylann: Fixed in 8b3909f: dropped the dead `catch let error as CastError { throw error }` branch. Confirmed via `rg 'CastError\.' Sources/Cast/Schema` (0 hits) — neither `SchemaGenerator.schema(for:)` nor `annotations(for:)` (which both delegate through `cached(_:)` → `build(_:)` / `ZeroSchemaDecoder.decode`) ever throw `CastError`. The remaining generic `catch` that wraps in `CastError.schemaGenerationFailed` is the only catch that ever runs, so simplified to that.

PR #103, opened 2026-05-01
fix(api): re-export Collections + JSONSchema from Cast (#74)

### LINKED ISSUE #74: @Castable consumers need 3 imports — re-export JSONSchema and Collections from Cast
## Problem

The README's Quick Start says `import Cast` is enough to use `@Castable`, but it isn't. The macro expands into:

```swift
static let castSchema: JSONSchema = .object(
    properties: OrderedDictionary(dictionaryLiteral: ...)
    ...
)
```

Both `JSONSchema` and `OrderedDictionary` need to be in scope at the call site, so consumers actually need:

```swift
import Cast
import Collections   // for OrderedDictionary
import JSONSchema    // for JSONSchema
```

Without the extra imports, the build fails with `cannot find type 'JSONSchema' in scope` *inside the macro expansion* — which is opaque to anyone who hasn't seen it before. Surfaced while writing the new `Examples/` subpackage (#73): every `@Castable`-using example needs both extra imports + extra dependencies in `Package.swift`.

## Fix

One-line change inside `Sources/Cast/API/Castable.swift`:

```swift
@_exported import Collections
@_exported import JSONSchema

@attached(member, names: named(castSchema), named(init))
@attached(extension, conformances: Castable, Decodable)
public macro Castable() = #externalMacro(module: "CastMacros", type: "CastableMacro")
```

After this, `import Cast` is genuinely sufficient. `Examples/Package.swift` can drop the `swift-json-schema` and `swift-collections` direct dependencies, and the `import Collections` / `import JSONSchema` lines disappear from every example.

Verified locally during the #73 work — it compiles and all examples build cleanly. Was scoped out of the Examples PR to keep that change additive-only.

## Acceptance

- [ ] `Sources/Cast/API/Castable.swift` adds the two `@_exported import`s.
- [ ] `Examples/Package.swift` no longer declares `swift-json-schema` or `swift-collections` as dependencies.
- [ ] Every `Examples/Sources/<Name>/main.swift` drops `import Collections` and `import JSONSchema`.
- [ ] README Quick Start unchanged — `import Cast` claim is now correct.
- [ ] `swift build` passes at repo root and inside `Examples/`.

## Related

- Discovered while implementing #73 (examples epic).

jaylann:
## Summary

The `@Castable` macro expansion references `JSONSchema` and `OrderedDictionary` directly, so consumers had to add three imports (`import Cast` + `import Collections` + `import JSONSchema`) instead of the one the README claims is sufficient. Without the extra imports, the build fails with `cannot find type 'JSONSchema' in scope` *inside* the macro expansion — which is opaque if you haven't seen it before.

This PR adds `@_exported import Collections` and `@_exported import JSONSchema` to `Sources/Cast/API/Castable.swift`, then drops the now-redundant package deps from `Examples/Package.swift` and the `import` lines from every example file.

## Changes

- `Sources/Cast/API/Castable.swift` — re-export `Collections` and `JSONSchema`.
- `Examples/Package.swift` — drop `swift-json-schema` and `swift-collections` from package + target deps.
- `Examples/Sources/*/main.swift` — drop `import Collections` and `import JSONSchema` lines.
- `Sources/Cast/Cast.docc/Examples/*.md` — regenerated from updated examples via `scripts/generate-example-docs.sh` (also added the previously-missing `ChatTemplates.md`).

## Test Plan

- [x] `swift build` at repo root passes
- [x] `swift test --filter` passes for macro/schema/wrapper/prompt/validator/CastEnum suites (no new failures; pre-existing MLX-runtime crash on `Index out of range` is unrelated to this change and reproduces on `stage`)
- [ ] `cd Examples && swift build` — verified by CI from the `Cast/` checkout (locally fails inside this worktree due to the documented `unknown package 'Cast'` path-package quirk for non-`Cast/`-named worktrees)

Closes #74

## Release Notes

* **New Features**
  * Collections and JSONSchema are now re-exported from the Cast module, allowing access to these libraries with a single `import Cast` statement.

* **Documentation**
  * Updated all examples to remove unused imports and improve code clarity.
  * Improved ChatTemplates documentation formatting for chat-template markers.

* **Tests**
  * Added test coverage to ensure re-exported dependencies are properly accessible to consumers.

### DIFF
--- a/Examples/Sources/ChatTemplates/main.swift
+++ b/Examples/Sources/ChatTemplates/main.swift
@@ -6,9 +6,7 @@
 // <start_of_turn>. No per-family code paths in Cast.
 
 import Cast
-import Collections
 import Foundation
-import JSONSchema
 
 @Castable
 struct Greeting {
--- a/Examples/Sources/ChatTemplates/main.swift
+++ b/Examples/Sources/ChatTemplates/main.swift
@@ -1,9 +1,9 @@
 // What this shows: the same prompt routed through five different model
 // families to demonstrate cross-family chat-template support. Cast hands a
 // flat "(system)\n\n(prompt)" string to MLXLMCommon's processor.prepare,
-// which applies each tokenizer's chat_template — Qwen <|im_start|>,
-// Llama <|begin_of_text|>, Mistral [INST], Phi <|user|>, Gemma
-// <start_of_turn>. No per-family code paths in Cast.
+// which applies each tokenizer's chat_template — Qwen `<|im_start|>`,
+// Llama `<|begin_of_text|>`, Mistral `[INST]`, Phi `<|user|>`, Gemma
+// `<start_of_turn>`. No per-family code paths in Cast.
 
 import Cast
 import Foundation
--- a/Sources/Cast/API/Castable.swift
+++ b/Sources/Cast/API/Castable.swift
@@ -1,3 +1,6 @@
+@_exported import Collections
+@_exported import JSONSchema
+
 /// Generates a JSON Schema and `Decodable` conformance for a struct, making it
 /// usable as the output type of ``CastModel/cast(_:as:system:config:didGenerate:)-1jybk``.
 ///
--- a/Sources/Cast/Cast.docc/Examples/ChatTemplates.md
+++ b/Sources/Cast/Cast.docc/Examples/ChatTemplates.md
@@ -3,9 +3,9 @@
 the same prompt routed through five different model
 families to demonstrate cross-family chat-template support. Cast hands a
 flat "(system)\n\n(prompt)" string to MLXLMCommon's processor.prepare,
-which applies each tokenizer's chat_template — Qwen <|im_start|>,
-Llama <|begin_of_text|>, Mistral [INST], Phi <|user|>, Gemma
-<start_of_turn>. No per-family code paths in Cast.
+which applies each tokenizer's chat_template — Qwen `<|im_start|>`,
+Llama `<|begin_of_text|>`, Mistral `[INST]`, Phi `<|user|>`, Gemma
+`<start_of_turn>`. No per-family code paths in Cast.
 
 ## Source
 
@@ -15,9 +15,9 @@ Full source: [Examples/Sources/ChatTemplates/main.swift](https://github.com/jayl
 // What this shows: the same prompt routed through five different model
 // families to demonstrate cross-family chat-template support. Cast hands a
 // flat "(system)\n\n(prompt)" string to MLXLMCommon's processor.prepare,
-// which applies each tokenizer's chat_template — Qwen <|im_start|>,
-// Llama <|begin_of_text|>, Mistral [INST], Phi <|user|>, Gemma
-// <start_of_turn>. No per-family code paths in Cast.
+// which applies each tokenizer's chat_template — Qwen `<|im_start|>`,
+// Llama `<|begin_of_text|>`, Mistral `[INST]`, Phi `<|user|>`, Gemma
+// `<start_of_turn>`. No per-family code paths in Cast.
 
 import Cast
 import Foundation
--- a/Sources/Cast/Cast.docc/Examples/ChatTemplates.md
+++ b/Sources/Cast/Cast.docc/Examples/ChatTemplates.md
@@ -20,9 +20,7 @@ Full source: [Examples/Sources/ChatTemplates/main.swift](https://github.com/jayl
 // <start_of_turn>. No per-family code paths in Cast.
 
 import Cast
-import Collections
 import Foundation
-import JSONSchema
 
 @Castable
 struct Greeting {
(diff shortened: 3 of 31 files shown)


Sources/Cast/API/Castable.swift
```diff
@@ -1,3 +1,6 @@
+@_exported import Collections
+@_exported import JSONSchema
```
jaylann, line 2: No test currently guards this single-import contract. Every file under `Tests/CastTests/` still does `import Cast` + `import Collections` + `import JSONSchema` (e.g. `CastBenchE2ETests.swift`, `CastStreamTests.swift`, `ChatTemplateTests.swift`, `ExtractTests.swift`, `SchemaExcludingTests.swift`), so if someone removes one of these `@_exported` lines in a future cleanup the test suite still compiles and the regression that motivated #74 silently re-emerges in consumer code only.

A minimal guard would be a single test file (or even a `Sources/Cast/Cast.docc/SingleImportSmoketest.swift`-style compile-only file) whose entire contents are:

```swift
import Cast

@Castable struct _SingleImportProbe { var x: String = "" }
```

Compilation alone is the assertion — if `@_exported` is ever dropped, this file fails to build. Without it, the only thing protecting users is `Examples/Sources/HelloCast` which is a slow CI signal at best.

Sources/Cast/API/Castable.swift
```diff
@@ -1,3 +1,6 @@
+@_exported import Collections
```
jaylann, line 1: Putting these re-exports at the top of `Castable.swift` works, but it means the implicit public-surface widening (every `import Cast` consumer now also gets `OrderedDictionary`, `Deque`, `JSONSchema`, etc.) is invisible from a file-name grep. A dedicated `Sources/Cast/API/Exports.swift` whose entire contents are the two `@_exported import` lines plus a one-line comment ("// These are re-exported because @Castable expansion references JSONSchema and OrderedDictionary directly. Removing either breaks consumers.") would:

1. Signal intent — anyone auditing the public surface sees `Exports.swift` and knows what to look for.
2. Prevent accidental deletion — the lines arent buried above an unrelated macro doc-comment that someone might rewrite without noticing them.
3. Make the diff a single-file add rather than mutating the canonical macro declaration file.

Not blocking, but with `@_exported` being an underscored attribute the future-proofing argument cuts both ways: if Apple ever stabilizes a replacement, the migration is one file, not a hunt across `Sources/Cast/API/`.

Sources/Cast/API/Castable.swift
jaylann: Fixed in f8bb327ab26314d29f9c535f335817d1550cfab5: added `Tests/CastTests/ReExportsTest.swift` that imports only `Cast` and uses `OrderedDictionary<String, Int>` + `JSONSchema.string()`. Compilation alone proves the re-export contract — removing either `@_exported` line will break this file.

Examples/Sources/ChatTemplates/main.swift
```diff
@@ -6,9 +6,7 @@
 // <start_of_turn>. No per-family code paths in Cast.
 
 import Cast
-import Collections
 import Foundation
```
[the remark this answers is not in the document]
jaylann: Fixed in f8bb327ab26314d29f9c535f335817d1550cfab5: `import Foundation` is at line 9 (transitively re-exports `exit` on Apple platforms), so the `exit(1)` call resolves cleanly. Verified with `CI=true swift build` — green.

Sources/Cast/Cast.docc/Examples/ChatTemplates.md
```diff
@@ -0,0 +1,70 @@
+# ChatTemplates
+
+the same prompt routed through five different model
+families to demonstrate cross-family chat-template support. Cast hands a
+flat "(system)\n\n(prompt)" string to MLXLMCommon's processor.prepare,
+which applies each tokenizer's chat_template — Qwen <|im_start|>,
+Llama <|begin_of_text|>, Mistral [INST], Phi <|user|>, Gemma
+<start_of_turn>. No per-family code paths in Cast.
```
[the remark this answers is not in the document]
jaylann: Fixed in f8bb327ab26314d29f9c535f335817d1550cfab5: wrapped `<|im_start|>`, `<|begin_of_text|>`, `[INST]`, `<|user|>`, and `<start_of_turn>` in backticks. Note: `Cast.docc/Examples/*.md` is generated from `Examples/Sources/<Name>/main.swift` by `scripts/generate-example-docs.sh` (CI has a drift gate), so the actual edit is in the example source file's `// What this shows:` header; the article was regenerated to match.

Sources/Cast/API/Castable.swift
jaylann: Keeping these in `Castable.swift` on purpose — co-locating the `@_exported` lines next to the `@Castable` macro that depends on them. The `CLAUDE.md` learning documents the constraint ("must stay at the top of `Castable.swift`") so a future cleanup won't accidentally split them. Happy to revisit if we ever add a second batch of re-exports unrelated to the macro.

PR #104, opened 2026-05-01
feat(prompt): surface schema-encoding failures as CastError (#85)

### LINKED ISSUE #85: PromptEngine.buildDefaultSystem silently substitutes [Schema encoding failed]
## Problem

`PromptEngine.buildDefaultSystem` (and the parallel `buildExtractionSystem` introduced in #77) wrap the JSON-schema encoding step in `try?` and, on failure, substitute the literal string `[Schema encoding failed]` into the system prompt:

```swift
if let data = try? encoder.encode(schema),
   let str = String(data: data, encoding: .utf8) {
    parts.append("JSON Schema:")
    parts.append(str)
} else {
    parts.append("[Schema encoding failed]")
}
```

The model then receives a system prompt with no schema and produces low-quality output, while the caller has no signal that anything went wrong. For an `extract()` call this is particularly bad — the whole point of the helper is structured extraction against a schema.

## Suggested fix

Surface the encoding failure as `CastError.schemaGenerationFailed` from the prompt builder (or validate at the call site, which already has the schema in hand). Either:

1. Make `buildPrompt` / `buildExtractionPrompt` `throws` and propagate the underlying `EncodingError`.
2. Add a `try` validation step at the `cast()` / `extract()` call site that pre-encodes the schema and throws a typed error before prompt construction.

## Context

Flagged in review of #77 ([comment](https://github.com/jaylann/Cast/pull/77#discussion)) as out of scope for that PR — `extract()` inherits the behaviour from the existing `buildDefaultSystem` rather than introducing it.

## Acceptance criteria

- [ ] Schema-encoding failure surfaces as a typed `CastError` instead of being silently substituted into the prompt.
- [ ] Tests cover the failure path (e.g. with a schema type whose encoder throws).
- [ ] No behaviour change on the happy path.

jaylann:
## Summary

`PromptEngine.buildDefaultSystem` and `buildExtractionSystem` wrapped the JSON-schema encoding step in `try?` and substituted the literal string `[Schema encoding failed]` into the system prompt on failure. The model received a prompt with no schema, produced low-quality output, and the caller had no signal anything went wrong.

This PR makes `buildPrompt` and `buildExtractionPrompt` `throws` and propagates the failure as `CastError.schemaGenerationFailed`. `buildClassificationPrompt` is unaffected — it doesn't encode a schema.

## Changes

- `Sources/Cast/Prompt/PromptEngine.swift`: `buildPrompt`/`buildExtractionPrompt` now `throws`; encoding failures throw `CastError.schemaGenerationFailed`.
- `Sources/Cast/API/CastModel+Generation.swift`, `CastModel+Extract.swift`, `CastModel+Stream.swift`, `Sources/Cast/Bench/Instrumentation.swift`: callers add `try` (already in `async throws` context).
- `Tests/CastTests/PromptEngineTests.swift`: existing tests adapted to the new throws shape.

## Test Plan

- [x] `swift test --filter CastTests.PromptEngineTests` passes (12/12)
- [x] `swift test` full suite — no new failures (a pre-existing Metal-related fatal error in MLX runtime tests is unaffected; matches CLAUDE.md `requiresMetal` notes)
- [x] `swift build` confirms callers wired

## Note on failure-path testing

`JSONSchema` from `swift-json-schema` is a plain value type; with a default `JSONEncoder`, encoding never naturally throws. Constructing a synthetic failing input would require either a custom encoder injection point or mocking the schema type — both would widen the API surface beyond what this fix needs. The architectural fix surfaces the error type cleanly for any future encoding bug; the existing 12 tests prove no behavior change on the happy path.

Closes #85

## Summary by CodeRabbit

* **Breaking Changes**
  * Prompt-construction APIs now surface errors when schema encoding fails (previously silent); callers must handle these exceptions.

* **Documentation**
  * Added v1.0 migration notes describing source-breaking behavior and recommended try/catch guidance.

* **Tests**
  * Expanded tests to verify error behavior for schema encoding edge cases (e.g., non-JSON-finite values).

### DIFF
--- a/Sources/Cast/Prompt/PromptEngine.swift
+++ b/Sources/Cast/Prompt/PromptEngine.swift
@@ -2,6 +2,20 @@ import Foundation
 import JSONSchema
 
 public enum PromptEngine {
+    /// Build a system + user prompt pair that embeds the JSON schema the
+    /// model must conform to. When `system` is supplied, it is returned
+    /// verbatim as the system message; otherwise a default system message
+    /// is synthesized from the schema and any per-field annotations.
+    ///
+    /// - Parameters:
+    ///   - userPrompt: The user-facing instruction returned as the user message.
+    ///   - schema: The JSON schema the output must conform to.
+    ///   - annotations: Optional per-field guidance (description, examples).
+    ///   - system: Optional override for the system message.
+    /// - Throws: ``CastError/schemaGenerationFailed(_:)`` if the schema
+    ///   cannot be JSON-encoded (e.g. it carries a non-finite `Double`
+    ///   constant such as `.nan` that `JSONEncoder` rejects under its
+    ///   default `nonConformingFloatEncodingStrategy`).
     public static func buildPrompt(
         userPrompt: String,
         schema: JSONSchema,
@@ -94,6 +108,10 @@ public enum PromptEngine {
     ///   - schema: The JSON schema the output must conform to.
     ///   - annotations: Optional per-field guidance (description, examples).
     ///   - system: Optional override for the system message.
+    /// - Throws: ``CastError/schemaGenerationFailed(_:)`` if the schema
+    ///   cannot be JSON-encoded (e.g. it carries a non-finite `Double`
+    ///   constant such as `.nan` that `JSONEncoder` rejects under its
+    ///   default `nonConformingFloatEncodingStrategy`).
     public static func buildExtractionPrompt(
         text: String,
         instruction: String,
--- a/Sources/Cast/Prompt/PromptEngine.swift
+++ b/Sources/Cast/Prompt/PromptEngine.swift
@@ -7,11 +7,11 @@ public enum PromptEngine {
         schema: JSONSchema,
         annotations: [String: FieldAnnotation] = [:],
         system: String? = nil
-    ) -> (system: String, user: String) {
+    ) throws -> (system: String, user: String) {
         let systemPrompt: String = if let system {
             system
         } else {
-            buildDefaultSystem(schema: schema, annotations: annotations)
+            try buildDefaultSystem(schema: schema, annotations: annotations)
         }
         return (system: systemPrompt, user: userPrompt)
     }
@@ -19,7 +19,7 @@ public enum PromptEngine {
     private static func buildDefaultSystem(
         schema: JSONSchema,
         annotations: [String: FieldAnnotation]
-    ) -> String {
+    ) throws -> String {
         var parts: [String] = []
 
         parts.append("You are a structured data extraction assistant.")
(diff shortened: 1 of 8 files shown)


Sources/Cast/Prompt/PromptEngine.swift
```diff
@@ -28,13 +28,19 @@ public enum PromptEngine {
...
-            parts.append(str)
-        } else {
-            parts.append("[Schema encoding failed]")
+        let data: Data
+        do {
+            data = try encoder.encode(schema)
+        } catch {
+            throw CastError.schemaGenerationFailed(
+                "Failed to encode JSON schema: \(error.localizedDescription)"
+            )
```
jaylann, line 51: The PR description states a synthetic failure-path test would require API widening, but a default `JSONEncoder` will throw on a `Double.nan` (its `nonConformingFloatEncodingStrategy` defaults to `.throw`), and `JSONSchema.NumberSchema` exposes `Double?` bounds (`minimum`, `maximum`, `exclusiveMinimum`, `exclusiveMaximum`, `multipleOf`). A test like `JSONSchema.number(minimum: .nan)` fed straight into `PromptEngine.buildPrompt` should reach this `throw` branch without any new public API. Worth adding so this branch isn't covered only by inspection.

Sources/Cast/Prompt/PromptEngine.swift
```diff
@@ -94,16 +100,20 @@ public enum PromptEngine {
         schema: JSONSchema,
         annotations: [String: FieldAnnotation] = [:],
         system: String? = nil
-    ) -> (system: String, user: String) {
-        let systemPrompt = system ?? buildExtractionSystem(schema: schema, annotations: annotations)
+    ) throws -> (system: String, user: String) {
+        let systemPrompt: String = if let system {
+            system
```
jaylann, line 123: `buildPrompt` and `buildExtractionPrompt` are now `throws`, but neither doc comment has a `- Throws:` clause. `buildPrompt` has no doc comment at all; `buildExtractionPrompt`'s existing comment (above, around lines 84-90) was left unchanged. The DocC site is the public surface for these functions — please add a `- Throws: ``CastError/schemaGenerationFailed(_:)`` if the schema cannot be JSON-encoded.` line on both so external callers know what error they're catching, and what its semantic is.

Sources/Cast/Prompt/PromptEngine.swift
```diff
@@ -7,19 +7,19 @@ public enum PromptEngine {
         schema: JSONSchema,
         annotations: [String: FieldAnnotation] = [:],
         system: String? = nil
-    ) -> (system: String, user: String) {
+    ) throws -> (system: String, user: String) {
```
jaylann, line 24: Adding `throws` to a `public static func` is a source-breaking change for any external caller using `PromptEngine.buildPrompt` / `buildExtractionPrompt` directly (the call sites inside Cast itself are fine — those got `try`). Pre-1.0 this is acceptable, but worth a one-liner in `MIGRATION.md` (or a `### Breaking changes` note in the PR body that gets surfaced in the eventual release notes) so anyone calling `PromptEngine` outside the convenience APIs sees a heads-up before the next tag.

Sources/Cast/Prompt/PromptEngine.swift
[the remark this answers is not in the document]
jaylann: Keeping inline — only two call sites and the new nan-based test (thread above) covers the throw branch without API widening.

jaylann: Fixed in 46ceec1: added `PromptEngineThrowsTests` using `JSONSchema.number(minimum: .nan)` to hit the throw branch via `JSONEncoder`'s default non-conforming-float policy.

Sources/Cast/Prompt/PromptEngine.swift
jaylann: Fixed in 46ceec1: added `- Throws:` clauses to `buildPrompt` (with new doc comment) and `buildExtractionPrompt`.

Sources/Cast/Prompt/PromptEngine.swift
jaylann: Fixed in 46ceec1: added a v1.0 `Source-breaking changes` section to MIGRATION.md noting both surfaces are now `throws` and what to catch.

Issue #105, opened 2026-05-01
Cast doesn't build for iOS — Sendable capture errors in CastModel+Lifecycle.swift


jaylann:
## Problem

Building `Cast` for an iOS target (e.g. as a SwiftPM dependency in an iOS app project) fails with three Swift 6 strict-concurrency errors in `Sources/Cast/API/CastModel+Lifecycle.swift`:

\`\`\`
Sources/Cast/API/CastModel+Lifecycle.swift:108:41: error: capture of 'didEnter' with non-Sendable type 'any NSObjectProtocol' in a '@Sendable' closure
Sources/Cast/API/CastModel+Lifecycle.swift:109:39: error: capture of 'willResign' with non-Sendable type 'any NSObjectProtocol' in a '@Sendable' closure
Sources/Cast/API/CastModel+Lifecycle.swift:110:36: error: capture of 'memWarn' with non-Sendable type 'any NSObjectProtocol' in a '@Sendable' closure
\`\`\`

The errors fire inside `enableBackgroundSafety()` where the `NotificationCenter.addObserver` return values (typed `any NSObjectProtocol`) are passed into the `_observers.withLock { dict in … }` closure (which is `@Sendable`).

## Repro

\`\`\`bash
xcodebuild -scheme Cast -destination 'generic/platform=iOS Simulator' build
\`\`\`

(macOS `swift build` does not trigger this — `enableBackgroundSafety()` is iOS-only and gated by `#if canImport(UIKit) && os(iOS)`.)

## Impact

Discovered while implementing #44 (SwiftUI demo). The iOS Xcode demo app (`Examples/iOS/CastDemo/`) cannot build until this is fixed. The macOS SwiftUI executable (`Examples/Sources/SwiftUIDemo/`) builds fine.

## Suggested fix

Wrap the three observer tokens in a small `@unchecked Sendable` box (similar to the `WeakBox` pattern already in this file), or use the typed `NSNotification.Name`-based observer API and cast/store directly. The observer tokens only need to outlive the lifetime of the model and be passed to `removeObserver(_:)` — they don't need cross-actor mutation.

Discovered: 2026-05-01.

Issue #41, opened 2026-03-20
Timeout and cancellation support


jaylann:
Graceful timeout and cancellation for generation.

- Configurable timeout per generation call
- Cancellation via Swift structured concurrency (Task cancellation)
- Distinguish timeout from user-requested abort
- Clean GPU state after cancellation

jaylann:
Shipped in 37ec693.

- New `CastConfiguration.timeout: Duration?` (nil = no timeout).
- Generation runs inside a `withThrowingTaskGroup` race; on deadline, throws `CastError.timedOut(partialOutput:)` and the inner sampler's `Task.isCancelled` poll exits cleanly.
- External `Task.cancel()` now throws `CastError.cancelled(partialOutput:)` carrying the bytes the model produced (previously cancellation surfaced as `CastError.decodingFailed`). Behavior change is documented in `MIGRATION.md` and the `Cancellation` example.
- `cleanupGPU()` runs on every error path including timeout/cancel.

PR #106, opened 2026-05-01
feat(examples): SwiftUI demo app — macOS + iOS (#44)

### LINKED ISSUE #44: Example iOS app with SwiftUI + streaming fields
Demo iOS app showing Cast in action.

- SwiftUI interface
- Load a model, enter a prompt
- Streaming fields populate in real-time as model generates
- Show multiple generation modes: cast, classify, extract

jaylann:
## Summary

Closes the long-standing #44 (deferred until streaming + cancellation shipped). Two demo surfaces:

- **macOS SwiftUI executable** at `Examples/Sources/SwiftUIDemo/main.swift` — built by Examples CI; runnable via `swift run SwiftUIDemo`. Quick `swift`-only smoke without Xcode.
- **iOS Xcode app** at `Examples/iOS/CastDemo/` — XcodeGen-driven (`project.yml` + sources, no committed `.xcodeproj`). Run `xcodegen generate` to materialize the project, open in Xcode, run on Simulator or device.

Both demonstrate:
- Loading a model with `CastModel.load(...)`
- Three modes via a `Picker`: `cast()` (Recipe), `classify()` (Sentiment), `extract()` (InvoiceFields)
- Live-streaming partial Recipe fields via `castStream()` with a progress bar
- Cancel button wired to `model.abortInFlight()` (#41)
- iOS demo additionally calls `enableBackgroundSafety()` on load

## Why two surfaces

- iOS apps can't ship as a Swift Package `.executableTarget` — they need an `.app` bundle, which only Xcode produces. A hand-written `project.pbxproj` is fragile, so the iOS demo uses XcodeGen (`brew install xcodegen` → `xcodegen generate`).
- The macOS `SwiftUIDemo` target gives Examples CI an automated build check on the SwiftUI + Cast wiring without needing xcodebuild infrastructure.

## Test Plan

- [x] `swift build` at repo root passes (Cast unchanged)
- [x] `scripts/generate-example-docs.sh` mirrors the new article (`Cast.docc/Examples/SwiftUIDemo.md`); kept generated, not committed (CI regenerates)
- [x] `xcodegen generate` succeeds on `Examples/iOS/CastDemo/project.yml`
- [ ] macOS Examples build: not verified locally — the path-package quirk in CLAUDE.md (worktree dir != "Cast" → SPM derives `'examples'` and rejects `package: "Cast"`). Trust `.github/workflows/examples.yml` to validate from the main `Cast/` checkout.
- [ ] iOS Xcode build: blocked by pre-existing Cast iOS Sendable issue — see #105. The demo code compiles, but `Cast` itself fails to build for iOS targets due to three `non-Sendable NSObjectProtocol` capture errors in `CastModel+Lifecycle.swift:108-110`. Filing #105 separately; this PR's iOS demo will build cleanly once #105 lands.
- [ ] Manual UX smoke on iOS Simulator (left for repo owner — model weights need a real device or Wi-Fi)

## Notes

- `Examples/iOS/CastDemo/.xcodeproj` is gitignored — regenerated from `project.yml` by XcodeGen.
- `Examples/Package.swift` keeps `JSONSchema`/`Collections` target deps on the new `SwiftUIDemo` target; PR #74 (re-exports) will sweep these repo-wide. If #74 lands first, this PR rebases with a trivial conflict resolution.
- iOS demo `DemoModels.swift` does NOT carry the explicit `import Collections` / `import JSONSchema` — when #74 lands they'd be redundant. The macOS executable keeps them for now since Examples CI builds today.

Closes #44

### DIFF
--- a/Examples/Sources/SwiftUIDemo/main.swift
+++ b/Examples/Sources/SwiftUIDemo/main.swift
@@ -1,0 +1,301 @@
+// What this shows: SwiftUI app demonstrating cast/classify/extract with live-streaming Recipe fields.
+
+#if os(macOS) || os(iOS)
+
+    import Cast
+    import Collections
+    import JSONSchema
+    import Observation
+    import SwiftUI
+
+    // MARK: - Demo Models
+
+    @Castable
+    struct Recipe {
+        @Description("Short, punchy title")
+        var title: String = ""
+
+        @MaxCount(8)
+        var ingredients: [String] = []
+
+        @CastRange(1 ... 60)
+        var prepMinutes: Int = 0
+    }
+
+    enum Sentiment: String, CastEnum, CaseIterable {
+        case positive, negative, neutral
+    }
+
+    @Castable
+    struct InvoiceFields {
+        @Description("Invoice number from the document")
+        var invoiceNumber: String = ""
+
+        @Description("Total amount in USD")
+        var totalUSD: Double = 0
+    }
+
+    // MARK: - View Model
+
+    enum DemoMode: String, CaseIterable, Identifiable {
+        case cast = "Cast"
+        case classify = "Classify"
+        case extract = "Extract"
+
+        var id: String {
+            rawValue
+        }
+    }
+
+    @MainActor
+    @Observable
+    final class DemoViewModel {
+        var modelID: String = "mlx-community/Llama-3.2-3B-Instruct-4bit"
+        var prompt: String = "Quick weeknight pasta in 20 minutes."
+        var sourceText: String = """
+        ACME Corp.
+        Invoice Number: INV-7421
+        Date: 2025-03-12
+        Total Due: $1,250.00 USD
+        """
+
+        var mode: DemoMode = .cast
+
+        var model: CastModel?
+        var isLoading: Bool = false
+        var isGenerating: Bool = false
+
+        var partialRecipe: Recipe.PartiallyGenerated?
+        var progress: Double = 0
+        var classification: Sentiment?
+        var extracted: InvoiceFields?
+        var errorMessage: String?
+
+        var generationTask: Task<Void, Never>?
+
+        var canGenerate: Bool {
+            model != nil && !isGenerating && !isLoading
+        }
+
+        func loadModel() {
+            guard !isLoading else { return }
+            isLoading = true
+            errorMessage = nil
+            let id = modelID
+            Task { [weak self] in
+                do {
+                    let loaded = try await CastModel.load(id)
+                    await MainActor.run {
+                        guard let self else { return }
+                        self.model = loaded
+                        self.isLoading = false
+                    }
+                } catch {
+                    await MainActor.run {
+                        guard let self else { return }
+                        self.errorMessage = "Load failed: \(error.localizedDescription)"
+                        self.isLoading = false
+                    }
+                }
+            }
+        }
+
+        func generate() {
+            guard let model, !isGenerating else { return }
+            errorMessage = nil
+            partialRecipe = nil
+            classification = nil
+            extracted = nil
+            progress = 0
+            isGenerating = true
+
+            let mode = self.mode
+            let prompt = self.prompt
+            let sourceText = self.sourceText
+
+            generationTask = Task { [weak self] in
+                defer {
+                    Task { @MainActor [weak self] in
+                        self?.isGenerating = false
+                    }
+                }
+
+                do {
+                    switch mode {
+                    case .cast:
+                        for try await partial in model.castStream(prompt, as: Recipe.self) {
+                            let snapshot = partial.value
+                            let pct = partial.progress
+                            await MainActor.run { [weak self] in
+                                self?.partialRecipe = snapshot
+                                self?.progress = pct
+                            }
+                        }
+                    case .classify:
+                        let result: Sentiment = try await model.classify(prompt)
+                        await MainActor.run { [weak self] in
+                            self?.classification = result
+                        }
+                    case .extract:
+                        let fields: InvoiceFields = try await model.extract(
+                            from: sourceText,
+                            instruction: prompt
+                        )
+                        await MainActor.run { [weak self] in
+                            self?.extracted = fields
+                        }
+                    }
+                } catch {
+                    await MainActor.run { [weak self] in
+                        self?.errorMessage = error.localizedDescription
+                    }
+                }
+            }
+        }
+
+        func cancel() {
+            model?.abortInFlight()
+            generationTask?.cancel()
+        }
+    }
+
+    // MARK: - Views
+
+    struct ContentView: View {
+        @State private var vm = DemoViewModel()
+
+        var body: some View {
+            Form {
+                Section("Model") {
+                    HStack {
+                        TextField("Model ID", text: $vm.modelID)
+                            .textFieldStyle(.roundedBorder)
+                            .disabled(vm.isLoading || vm.model != nil)
+                        Button(vm.model == nil ? "Load" : "Loaded") {
+                            vm.loadModel()
+                        }
+                        .disabled(vm.isLoading || vm.model != nil)
+                        if vm.isLoading {
+                            ProgressView().controlSize(.small)
+                        }
+                    }
+                }
+
+                Section("Mode") {
+                    Picker("Mode", selection: $vm.mode) {
+                        ForEach(DemoMode.allCases) { m in
+                            Text(m.rawValue).tag(m)
+                        }
+                    }
+                    .pickerStyle(.segmented)
+                }
+
+                Section(vm.mode == .extract ? "Instruction" : "Prompt") {
+                    TextField("Prompt", text: $vm.prompt, axis: .vertical)
+                        .lineLimit(2 ... 5)
+                        .textFieldStyle(.roundedBorder)
+                }
+
+                if vm.mode == .extract {
+                    Section("Source Text") {
+                        TextEditor(text: $vm.sourceText)
+                            .font(.body.monospaced())
+                            .frame(minHeight: 100)
+                    }
+                }
+
+                Section("Run") {
+                    HStack {
+                        Button("Generate") { vm.generate() }
+                            .disabled(!vm.canGenerate)
+                        Button("Cancel") { vm.cancel() }
+                            .disabled(!vm.isGenerating)
+                        if vm.isGenerating {
+                            ProgressView().controlSize(.small)
+                        }
+                    }
+                }
+
+                Section("Output") {
+                    OutputView(vm: vm)
+                }
+
+                if let error = vm.errorMessage {
+                    Section {
+                        Text(error)
+                            .foregroundStyle(.red)
+                            .font(.callout)
+                    }
+                }
+            }
+            .padding()
+            .frame(minWidth: 600, minHeight: 500)
+        }
+    }
+
+    struct OutputView: View {
+        let vm: DemoViewModel
+
+        var body: some View {
+            switch vm.mode {
+            case .cast:
+                VStack(alignment: .leading, spacing: 8) {
+                    ProgressView(value: vm.progress)
+                    LabeledContent("Title") {
+                        Text(vm.partialRecipe?.title ?? "—")
+                            .foregroundStyle(vm.partialRecipe?.title == nil ? .secondary : .primary)
+                    }
+                    LabeledContent("Prep Minutes") {
+                        Text(vm.partialRecipe?.prepMinutes.map { "\($0) min" } ?? "—")
+                            .foregroundStyle(vm.partialRecipe?.prepMinutes == nil ? .secondary : .primary)
+                    }
+                    LabeledContent("Ingredients") {
+                        if let ingredients = vm.partialRecipe?.ingredients, !ingredients.isEmpty {
+                            VStack(alignment: .trailing) {
+                                ForEach(Array(ingredients.enumerated()), id: \.offset) { _, item in
+                                    Text(item ?? "—")
+                                }
+                            }
+                        } else {
+                            Text("—").foregroundStyle(.secondary)
+                        }
+                    }
+                }
+            case .classify:
+                if let result = vm.classification {
+                    LabeledContent("Sentiment", value: result.rawValue)
+                } else {
+                    Text("—").foregroundStyle(.secondary)
+                }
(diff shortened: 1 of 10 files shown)


Examples/iOS/CastDemo/project.yml
```diff
@@ -0,0 +1,30 @@
...
+    settings:
+      base:
+        SWIFT_VERSION: "6.0"
+        IPHONEOS_DEPLOYMENT_TARGET: "17.0"
+        TARGETED_DEVICE_FAMILY: "1,2"
+        INFOPLIST_KEY_UILaunchScreen_Generation: YES
+        INFOPLIST_KEY_UIApplicationSceneManifest_Generation: YES
+        INFOPLIST_KEY_UISupportedInterfaceOrientations_iPad: "UIInterfaceOrientationPortrait UIInterfaceOrientationPortraitUpsideDown UIInterfaceOrientationLandscapeLeft UIInterfaceOrientationLandscapeRight"
+        INFOPLIST_KEY_UISupportedInterfaceOrientations_iPhone: "UIInterfaceOrientationPortrait UIInterfaceOrientationLandscapeLeft UIInterfaceOrientationLandscapeRight"
+        INFOPLIST_KEY_NSAppTransportSecurity: ""
```
jaylann: `INFOPLIST_KEY_NSAppTransportSecurity: ""` is not valid — the underlying `NSAppTransportSecurity` Info.plist key takes a **dictionary** value, not a string. With `GENERATE_INFOPLIST_FILE=YES`, Xcode will either reject this at plist-merge time or silently emit an empty `<dict/>` that overrides the secure default and could allow arbitrary loads on some toolchain versions.

Hugging Face downloads are HTTPS-only and the iOS default ATS policy already permits them, so the simplest fix is to **delete this line entirely**. If a future model host genuinely requires non-HTTPS, set the dict explicitly via a custom `Info.plist` (XcodeGen `infoPlist:` key) rather than the build-setting shortcut, e.g. `NSAllowsArbitraryLoads: true`.

Examples/iOS/CastDemo/Sources/ContentView.swift
```diff
@@ -0,0 +1,247 @@
...
+                    loaded.enableBackgroundSafety()
+                }
+            } catch {
+                await MainActor.run {
+                    guard let self else { return }
+                    self.errorMessage = "Load failed: \(error.localizedDescription)"
+                    self.isLoading = false
+                }
+            }
+        }
```
jaylann, line 66: Concurrency smell flagged in the agent's own learning: `DemoViewModel` is `@MainActor`, so this unstructured `Task { ... }` already inherits MainActor isolation in Swift 6 — the inner `await MainActor.run { ... }` is a redundant double-hop and is exactly the kind of pattern that **builds clean on macOS but can flush out `@Sendable` errors only when compiled for iOS** (the agent flagged this in their notes).

Recommend collapsing to:

```swift
Task { [weak self] in
    guard let self else { return }
    do {
        let loaded = try await CastModel.load(self.modelID)
        self.model = loaded
        self.isLoading = false
        loaded.enableBackgroundSafety()
    } catch {
        self.errorMessage = "Load failed: \(error.localizedDescription)"
        self.isLoading = false
    }
}
```

The same pattern repeats in `generate()` (line 84) and `Examples/Sources/SwiftUIDemo/main.swift` (line 85, 116) — same fix applies.

Examples/iOS/CastDemo/Sources/ContentView.swift
```diff
@@ -0,0 +1,247 @@
...
+        let mode = self.mode
+        let prompt = self.prompt
+        let sourceText = self.sourceText
+
+        generationTask = Task { [weak self] in
+            defer {
+                Task { @MainActor [weak self] in
+                    self?.isGenerating = false
+                }
+            }
```
jaylann: `defer { Task { @MainActor in self?.isGenerating = false } }` opens a UX race: the `defer` body returns synchronously after enqueueing an unstructured Task, so when the `for try await` loop or the awaited `classify`/`extract` finishes (or throws), `isGenerating` stays `true` for at least one MainActor hop. If the user taps **Generate** in that window the `\!isGenerating` guard rejects them, which on a fast model + slow scheduler can happen visibly often.

Worse, if the parent Task is cancelled (your `vm.cancel()` or view tear-down), the `defer`-spawned child Task is unstructured and may not run reliably.

Since the enclosing `Task` already inherits `@MainActor` from the `@MainActor` view-model, just write:

```swift
generationTask = Task { [weak self] in
    defer { self?.isGenerating = false }
    do { ... } catch { ... }
}
```

No nested Task, no race. Same fix applies to `Examples/Sources/SwiftUIDemo/main.swift` line 117–119.

Examples/iOS/CastDemo/Sources/ContentView.swift
```diff
@@ -0,0 +1,247 @@
...
+                    self?.errorMessage = error.localizedDescription
+                }
+            }
+        }
+    }
+
+    func cancel() {
+        model?.abortInFlight()
+        generationTask?.cancel()
+    }
```
jaylann, line 117: `abortInFlight()` cancels **every in-flight call** registered on the model (per `CastModel+Lifecycle.swift`), and `castStream`'s `continuation.onTermination` already triggers task cancellation when the consuming Task is cancelled (per `CastModel+Stream.swift`). Calling both is redundant and the documentation is more nuanced than this Cancel button suggests:

- For **stream** mode (`cast`), the canonical cancel path is `generationTask?.cancel()` — the stream's `onTermination` propagates cancellation to the producer Task and yields `CastError.cancelled(partialOutput:)`.
- For **non-stream** modes (`classify`, `extract`), `abortInFlight()` *or* `Task.cancel()` works — the registry-based cancel is what surfaces `.cancelled` for those.

Since both branches go through this same Cancel handler, calling both is defensible — but the demo would teach the API better if it just used `generationTask?.cancel()` and a comment explaining cooperative cancellation. As written, a reader copies this pattern and will end up with `abortInFlight` calls that cancel **other in-flight generations they didn't intend to** (relevant once a real app has multiple Tasks per model).

Same comment applies to `Examples/Sources/SwiftUIDemo/main.swift` lines 156-159.

Examples/iOS/CastDemo/Sources/ContentView.swift
```diff
@@ -0,0 +1,247 @@
...
+
+                if let error = vm.errorMessage {
+                    Section {
+                        Text(error)
+                            .foregroundStyle(.red)
+                            .font(.callout)
+                    }
+                }
+            }
+            .navigationTitle("Cast Demo")
```
jaylann, line 189: `ContentView` has no `.onDisappear { vm.cancel() }` hook. If the user navigates away (push a different screen, swipe back to Springboard mid-stream, or in a tab-based app switches tab), the `generationTask` keeps running, holds the GPU, and writes `partialRecipe` snapshots into a view-model nobody is observing.

On iOS this is partially mitigated because `enableBackgroundSafety()` cancels on `didEnterBackground` — but **navigation away while the app is still foregrounded does NOT trigger that notification**. The GPU stays pinned and the user just sees the app feel sluggish. On macOS `enableBackgroundSafety()` is a no-op so this is even worse.

Add:

```swift
.onDisappear { vm.cancel() }
```

to the `Form` (after `.navigationTitle(...)`), and the same to the macOS `ContentView` in `Examples/Sources/SwiftUIDemo/main.swift`.

Examples/iOS/CastDemo/project.yml
jaylann: Fixed in 7e85482: removed invalid `INFOPLIST_KEY_NSAppTransportSecurity: ""` from project.yml — HF downloads are HTTPS so default ATS works.

Examples/iOS/CastDemo/Sources/ContentView.swift
jaylann: Fixed in 7e85482: dropped the inner `MainActor.run` hops — Task inherits MainActor isolation from the `@MainActor`-isolated `DemoViewModel` so we mutate state directly.

Examples/iOS/CastDemo/Sources/ContentView.swift
jaylann: Fixed in 7e85482: collapsed to `defer { self?.isGenerating = false }` — runs synchronously on MainActor in the same turn, no race window, fires reliably on cancellation.

Examples/iOS/CastDemo/Sources/ContentView.swift
jaylann: Fixed in 7e85482: `cancel()` now only calls `generationTask?.cancel()`. `castStream`'s `continuation.onTermination` already cancels the underlying generation; `abortInFlight()` was overbroad.

Examples/iOS/CastDemo/Sources/ContentView.swift
jaylann: Fixed in 7e85482: added `.onDisappear { vm.cancel() }` to the root `NavigationStack`'s view body.

Examples/iOS/CastDemo/Sources/DemoModels.swift
```diff
@@ -0,0 +1,26 @@
+import Cast
```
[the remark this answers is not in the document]
jaylann: Fixed in 7e85482: added `import JSONSchema` and `import Collections` to `DemoModels.swift` so the macro expansion compiles before #103 lands.

Examples/Sources/SwiftUIDemo/main.swift
```diff
@@ -0,0 +1,301 @@
...
+                    }
+                    LabeledContent("Prep Minutes") {
+                        Text(vm.partialRecipe?.prepMinutes.map { "\($0) min" } ?? "—")
+                            .foregroundStyle(vm.partialRecipe?.prepMinutes == nil ? .secondary : .primary)
+                    }
+                    LabeledContent("Ingredients") {
+                        if let ingredients = vm.partialRecipe?.ingredients, !ingredients.isEmpty {
+                            VStack(alignment: .trailing) {
+                                ForEach(Array(ingredients.enumerated()), id: \.offset) { _, item in
+                                    Text(item ?? "—")
```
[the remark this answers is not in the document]
jaylann: Fixed in 7e85482: replaced `Text(item ?? "—")` with `Text(item)` — `Recipe.PartiallyGenerated.ingredients` is `[String]?` with non-optional elements.

Sources/Cast/Cast.docc/Cast.md
```diff
@@ -50,6 +50,7 @@ invalid tokens during decoding — so the model can only emit JSON that matches.
 - <doc:PrepareWarmup>
 - <doc:Cancellation>
 - <doc:Streaming>
+- <doc:SwiftUIDemo>
```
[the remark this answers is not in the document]
jaylann: Fixed in 7e85482: ran `scripts/generate-example-docs.sh` to generate `Sources/Cast/Cast.docc/Examples/SwiftUIDemo.md` from the `// What this shows:` header in the example source.

Examples/iOS/CastDemo/Sources/DemoModels.swift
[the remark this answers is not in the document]
jaylann: Fixed in 7e85482: added `import JSONSchema` and `import Collections` to `DemoModels.swift` (defensive, no harm after #103 lands).

Examples/iOS/CastDemo/Sources/ContentView.swift
```diff
@@ -0,0 +1,247 @@
...
+                }
+                LabeledContent("Prep Minutes") {
+                    Text(vm.partialRecipe?.prepMinutes.map { "\($0) min" } ?? "—")
+                        .foregroundStyle(vm.partialRecipe?.prepMinutes == nil ? .secondary : .primary)
+                }
+                if let ingredients = vm.partialRecipe?.ingredients, !ingredients.isEmpty {
+                    LabeledContent("Ingredients") {
+                        VStack(alignment: .trailing) {
+                            ForEach(Array(ingredients.enumerated()), id: \.offset) { _, item in
+                                Text(item ?? "—")
```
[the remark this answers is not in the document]
jaylann: Fixed in 7e85482: replaced `Text(item ?? "—")` with `Text(item)` in `ContentView.swift:219`.

Examples/Sources/SwiftUIDemo/main.swift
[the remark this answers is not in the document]
jaylann: Fixed in 7e85482: replaced `Text(item ?? "—")` with `Text(item)` in `Examples/Sources/SwiftUIDemo/main.swift:255`.

Examples/iOS/CastDemo/project.yml
```diff
@@ -0,0 +1,30 @@
...
+  CastDemo:
+    type: application
+    platform: iOS
+    deploymentTarget: "17.0"
+    sources:
+      - path: Sources
+    dependencies:
+      - package: Cast
+        product: Cast
+    settings:
```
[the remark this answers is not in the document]
jaylann: Fixed in 7e85482: added `swift-json-schema` (petrukha-ivan/2.0.2) and `swift-collections` (apple/1.3.0) packages to `project.yml` and wired `JSONSchema` + `Collections` products into the `CastDemo` target.

Examples/iOS/CastDemo/project.yml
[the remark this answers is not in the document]
jaylann: Fixed in 7e85482: removed the invalid `INFOPLIST_KEY_NSAppTransportSecurity: ""` line entirely (HF downloads are HTTPS — default ATS works).

Examples/iOS/CastDemo/README.md
```diff
@@ -0,0 +1,36 @@
+# CastDemo (iOS)
+
+A SwiftUI iOS demo of the [Cast](https://github.com/jaylann/Cast) Swift package. Demonstrates the three primary modes — `cast()`, `classify()`, and `extract()` — with live-streaming partial results.
+
+## Prerequisites
+
+- Xcode 15+ (Swift 6.0)
```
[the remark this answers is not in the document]
jaylann: Fixed in 7e85482: README now reads `Xcode 16+ (or Xcode 15 with an external Swift 6 toolchain)`.

PR #107, opened 2026-05-01
fix(stage): minor cleanup batch (#97, #98, #99, #100)

### LINKED ISSUE #97: docs(repo): add CHANGELOG.md and SUPPORT.md
The repo has `CONTRIBUTING.md`, `SECURITY.md`, `CODE_OF_CONDUCT.md`, `LICENSE`, `NOTICE`, `MIGRATION.md` — but two standard OSS files are missing:

- **`CHANGELOG.md`** — synthesizes `release.yml`'s auto-generated GitHub Release notes in one git-tracked place. Use Keep-a-Changelog format. The first entry should be the `0.1.0` tag.
- **`SUPPORT.md`** — points users to GitHub Issues for bug reports / feature requests, and (if enabled) GitHub Discussions for usage questions. Mirrors the GitHub-recommended template.

## Verification

- Both files exist at repo root.
- `gh repo view --web` "Security" / "About" sidebars show the new files.
- Next release (`release.yml` run) updates `CHANGELOG.md` either by hand or via a follow-up automation issue.

### LINKED ISSUE #98: refactor(api): add top-of-file rationale comment for each CastModel+* extension
`Sources/Cast/API/` has six `CastModel+*.swift` extensions:

| File | Lines | Public methods |
|---|---|---|
| `CastModel+Generation.swift` | 336 | `cast`, overloads |
| `CastModel+Stream.swift` | 288 | `castStream` |
| `CastModel+Lifecycle.swift` | 148 | load/unload, observers |
| `CastModel+Extract.swift` | 62 | `extract` |
| `CastModel+Timeout.swift` | 60 | `withTimeout` helpers |
| `CastModel+GPUSafety.swift` | 42 | GPU lifecycle helpers |

The split is idiomatic for `CastModel.swift` (105 LoC core) but the *intent* of each split — what conceptual surface each file owns — isn't documented. A new contributor adding `model.classify` has no rule for picking the right file.

## Suggested fix

Add a 1–3 line top-of-file comment to each `CastModel+*.swift` that states the surface it owns and the criterion for adding new methods. No code change. Optionally fold `+Timeout.swift` into `+Generation.swift` if the surface stays small after audit.

## Verification

- Each `CastModel+*.swift` has a header comment.
- A short contributor note in `.claude/rules/parallel-work.md` (or a new rules file) summarizes the split.

### LINKED ISSUE #99: refactor(schema): replace JSONSchema.excluding regex round-trip with custom Encoder
`Sources/Cast/Schema/JSONSchema+Excluding.swift` implements `excluding(fields:)` by:

1. Encode the `JSONSchema` to JSON.
2. Round-trip through `JSONSerialization` to a `[String: Any]` dict.
3. Filter the desired keys.
4. Encode back to JSON.
5. Decode back into `JSONSchema`.

This works today but is fragile to upstream `JSONSchema` library changes (any new key or shape can silently survive or be dropped) and to corner-case property names. The transient JSON string is also wasteful for large schemas.

## Suggested fix

When the `JSONSchema` package exposes a public mutable AST or a custom-Encoder hook, replace the round-trip with a structural walk that preserves type safety. In the meantime, add a regression test that locks the current behavior so future JSONSchema bumps are caught.

## Verification

- `Tests/CastTests/SchemaExcludingTests.swift` already covers the public behavior; no behavioral change in this issue.
- Replacement implementation must keep the existing tests green.
- Track the upstream library issue (if any) for the AST-mutation feature.

## Related

- This is one of the topics in the ADR backfill (#95) — ADR-0007 should describe the fragility and the migration plan.

### LINKED ISSUE #100: macro: promote unknownNonPrimitiveType warning to error (or emit fix-it)
`Sources/CastMacros/CastableDiagnostic.swift:57-62` emits `unknownNonPrimitiveType` as a **warning**, not an error:

```swift
var severity: DiagnosticSeverity {
    switch self {
    case .unknownNonPrimitiveType: .warning
    default: .error
    }
}
```

Because expansion proceeds on warnings (per the early-return guard, which now correctly filters on `severity == .error`), the consumer's downstream compile fails with a confusing `<TypeName>.PartiallyGenerated.PartiallyGenerated?` shape — a developer who missed the warning sees only the cryptic compile error.

Examples that trigger it: `Date`, `URL`, `UUID`, `Data`, `Decimal` fields.

## Options

1. **Promote to error.** Strictest. Breaks anyone currently relying on the warning. Force them to wrap or pre-convert. Recommended if there's no real workflow that needs the warning to be permissive.
2. **Keep warning + add a fix-it.** Suggest "Wrap in a small `@Castable` struct" or "Pre-convert to ISO-8601 String for Date". Less disruptive but still helpful.
3. **Add an opt-in `@Castable(strict: true)`** that promotes to error in opt-in mode.

Recommendation: (1) for v1.0, paired with a one-line MIGRATION note.

## Verification

- New macro test asserts the diagnostic *severity* changes from `.warning` to `.error`.
- Updated `CastDiagnosticTests.swift` exercises the new severity.
- A doc note in `Sources/Cast/Cast.docc/Examples/PropertyWrappersTour.md` (or wherever Foundation-type examples live) explains the supported wrapping pattern.

jaylann:
## Summary

Four small "minor" issues batched into one PR ahead of v1.0 polish:

- **#97** — Add `CHANGELOG.md` (Keep-a-Changelog format, 0.1.0 backfill) and `SUPPORT.md` (GitHub-recommended template). Both files are missing from the repo and are standard for OSS projects on the GitHub Community Standards page.
- **#98** — Add a 2–3 line `// File rationale:` header to each of the six `Sources/Cast/API/CastModel+*.swift` extensions stating the surface it owns and what doesn't belong. Add a "Picking a CastModel+*.swift file" table to `CONTRIBUTING.md`. No code change. Decision: keep `+Timeout.swift` separate (it's used by both `+Generation` and `+Stream`; folding would re-introduce coupling).
- **#99** — Add six regression tests to `SchemaExcludingTests` that lock the structural behavior of the JSON-string round-trip in `JSONSchema.excluding(fields:)` so future `swift-json-schema` upstream bumps surface as test failures, not silent shape drift. Implementation rewrite is deferred per the issue body until upstream exposes a public AST hook.
- **#100** — Promote `unknownNonPrimitiveType` from warning to **error**. Bare `Date`/`URL`/`UUID`/`Data`/`Decimal` fields used to emit a warning and let expansion proceed, producing a cryptic downstream `<TypeName>.PartiallyGenerated.PartiallyGenerated?` compile error. Now surfaced at the macro site. New `MIGRATION.md §3 — Foundation types` documents the two supported workarounds.

⚠️ **Breaking** — `@Castable struct Foo { var when: Date }` no longer compiles. Migrate per `MIGRATION.md §3`.

## Test Plan

- [x] `swift build` — succeeds (only pre-existing deprecation warnings in `MLX.setErrorHandler`).
- [x] `swift test --filter SchemaExcludingTests` — 14 tests pass (8 existing + 6 new).
- [x] `swift test --filter CastMacroTests` — 26 tests pass; the renamed `unknownNonPrimitiveType` test asserts `.error` severity and a no-expansion outcome.
- [x] `CI=true swift test` — full suite (206 tests) passes.

Closes #97
Closes #98
Closes #99
Closes #100

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ GitHub Releases (auto-generated from squash-merged PR titles) remain the canonic
 
 ## [Unreleased]
 
+### Added
+
+- `CastModel.extract(from:as:instruction:…)` — extraction-flavored entrypoint.
+- `CastModel.castStream(...)` — `AsyncSequence` of `PartialResult<T>` for streaming partial decoding.
+
 ## [0.1.0] - 2026-04-30
 
 First public release. Cast ships as an Apache-2.0 Swift Package: type-safe structured output from any local LLM on Apple Silicon via constrained decoding.
@@ -15,12 +20,12 @@ First public release. Cast ships as an Apache-2.0 Swift Package: type-safe struc
 ### Added
 
 - `@Castable` macro that synthesizes JSON Schema, `Decodable`, and a `PartiallyGenerated` mirror from an annotated struct.
-- `CastModel` with `cast`, `castJSON`, `classify`, `extract`, and `castStream` entrypoints; `init(wrapping:configuration:)` for caller-managed `ModelContainer` lifetimes.
+- `CastModel` with `cast`, `castJSON`, and `classify` entrypoints; `init(wrapping:configuration:)` for caller-managed `ModelContainer` lifetimes.
 - Property wrappers covering schema constraints: `@MaxLength`/`@MinLength`, `@Count`/`@MaxCount`/`@MinCount`, `@CastRange`, `@Precision`, `@Pattern`, `@OneOf`, `@Description`, `@Examples`, `@Nullable`, `@DefaultValue`, `@Validator`.
 - `JSONSchema.excluding(fields:)` for dynamic schema modification at call time.
 - `CastModel.prepare(_:)` warm-up to pay grammar-compilation cost upfront.
 - `didGenerate` callback for cooperative cancellation; `abortInFlight()` plus iOS background-safety hooks (`enableBackgroundSafety()` / `disableBackgroundSafety()`).
-- DocC documentation site auto-generated from runnable `Examples/` snippets (9 example targets).
+- DocC documentation site auto-generated from runnable `Examples/` snippets.
 - README, MIGRATION guide, security policy, and contributing guide.
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,44 @@
+# Changelog
+
+All notable changes to Cast are documented here.
+
+The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+GitHub Releases (auto-generated from squash-merged PR titles) remain the canonical source per release; this file is the human-readable summary kept under git so users can read it offline.
+
+## [Unreleased]
+
+## [0.1.0] - 2026-04-30
+
+First public release. Cast ships as an Apache-2.0 Swift Package: type-safe structured output from any local LLM on Apple Silicon via constrained decoding.
+
+### Added
+
+- `@Castable` macro that synthesizes JSON Schema, `Decodable`, and a `PartiallyGenerated` mirror from an annotated struct.
+- `CastModel` with `cast`, `castJSON`, `classify`, `extract`, and `castStream` entrypoints; `init(wrapping:configuration:)` for caller-managed `ModelContainer` lifetimes.
+- Property wrappers covering schema constraints: `@MaxLength`/`@MinLength`, `@Count`/`@MaxCount`/`@MinCount`, `@CastRange`, `@Precision`, `@Pattern`, `@OneOf`, `@Description`, `@Examples`, `@Nullable`, `@DefaultValue`, `@Validator`.
+- `JSONSchema.excluding(fields:)` for dynamic schema modification at call time.
+- `CastModel.prepare(_:)` warm-up to pay grammar-compilation cost upfront.
+- `didGenerate` callback for cooperative cancellation; `abortInFlight()` plus iOS background-safety hooks (`enableBackgroundSafety()` / `disableBackgroundSafety()`).
+- DocC documentation site auto-generated from runnable `Examples/` snippets (9 example targets).
+- README, MIGRATION guide, security policy, and contributing guide.
+
+### Changed
+
+- License unified to **Apache-2.0** across the package, including vendored upstreams (`petrukha-ivan/mlx-swift-structured` and `mlc-ai/xgrammar`); attribution preserved in `NOTICE`.
+- Repository moved to a staged release model: `stage` is the integration branch and PR target, `main` is release-only and updated by `.github/workflows/release.yml`.
+
+### Fixed
+
+- `@Castable` now synthesizes `Decodable` conformance for the consumer.
+- `MLXStructured/LICENSE` is excluded from the SPM target so the build no longer warns about an unhandled file.
+- xgrammar pinned to `v0.1.34` with `tvm_ffi` python bindings excluded.
+
+### Infrastructure
+
+- Manual release workflow (`release.yml`) with semver validation, tag uniqueness check, and CI-green gate before fast-forwarding `stage` → `main`.
+- `pre-public-check.sh` safety scanner (secrets, gitleaks, fork-PR audit) for every visibility / history change.
+- CI skips MLX runtime tests on `macos-15` runners via the `.requiresMetal` Swift Testing trait.
+
+[Unreleased]: https://github.com/jaylann/Cast/compare/v0.1.0...HEAD
+[0.1.0]: https://github.com/jaylann/Cast/releases/tag/v0.1.0
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,7 +68,7 @@ Use descriptive branch names:
 | `CastModel+Timeout.swift` | internal cross-isolation glue (`withGenerationTimeout`, `withInFlightRegistration`) | new public API |
 | `CastModel+GPUSafety.swift` | Metal/MLX lifecycle plumbing (`cleanupGPU`, global error handler) | code that doesn't touch `MLX.Stream` / `Memory` |
 
-If a new method doesn't fit any existing surface, add a new `CastModel+<Surface>.swift` rather than overloading one of the above. The first line of every such file should be a 2–3 line `// File rationale: …` comment that states what it owns and what it doesn't.
+If a new method doesn't fit any existing surface, add a new `CastModel+<Surface>.swift` rather than overloading one of the above. The first line of every such file should be a short header `// File rationale: …` comment that states what it owns and what it doesn't.
 
 ## Pull Request Process
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,6 +55,21 @@ Use descriptive branch names:
 - Use Swift Testing framework (`import Testing`)
 - Macro changes need expansion tests using `assertMacroExpansion`
 
+## Picking a `CastModel+*.swift` file
+
+`Sources/Cast/API/` splits the `CastModel` surface across six extension files. Use the table below when you add a new method so it lands in the right one — the file each method lives in is its surface contract.
+
+| File | Owns | Don't put here |
+|---|---|---|
+| `CastModel+Generation.swift` | every blocking generation entrypoint: `cast`, `castJSON`, `classify` | streaming, extraction-flavored prompts |
+| `CastModel+Stream.swift` | `castStream` and partial-decode helpers | blocking generation |
+| `CastModel+Extract.swift` | `extract(from:as:instruction:…)` and future extraction-shaped APIs | unrelated prompt templates |
+| `CastModel+Lifecycle.swift` | `abortInFlight()`, iOS background-safety hooks | model load/unload (the core `CastModel` type) |
+| `CastModel+Timeout.swift` | internal cross-isolation glue (`withGenerationTimeout`, `withInFlightRegistration`) | new public API |
+| `CastModel+GPUSafety.swift` | Metal/MLX lifecycle plumbing (`cleanupGPU`, global error handler) | code that doesn't touch `MLX.Stream` / `Memory` |
+
+If a new method doesn't fit any existing surface, add a new `CastModel+<Surface>.swift` rather than overloading one of the above. The first line of every such file should be a 2–3 line `// File rationale: …` comment that states what it owns and what it doesn't.
+
 ## Pull Request Process
 
 1. Fork the repository
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -21,7 +21,7 @@ For each existing type you want to migrate, walk these in order:
 
 - [ ] **Is it a `struct`?** Classes, actors, and `enum` (other than raw-value via `CastEnum`) are not supported by `@Castable`. The macro emits a `requiresStruct` diagnostic for non-structs.
 - [ ] **Are all stored properties `Sendable`?** Closures, `class` references, `weak` refs, `NSObject` subclasses, and most `AnyObject`-typed values are not.
-- [ ] **Are all stored properties `Decodable`?** Primitives (`String`, `Int`, `Double`, `Float`, `Bool`), arrays of `Decodable`, optionals of `Decodable`, and other `@Castable` types are fine. Custom types with `init(from:)` are fine.
+- [ ] **Are all stored properties `Decodable`?** Primitives (`String`, `Int`, `Double`, `Float`, `Bool`), arrays of `Decodable`, optionals of `Decodable`, and other `@Castable` types are fine. Custom *user-defined* types with `init(from:)` are fine; Foundation value types (`Date`, `URL`, `UUID`, `Data`, `Decimal`) are covered separately below.
 - [ ] **Are nested types also `@Castable`** (or otherwise `Decodable`)?
 - [ ] **No bare Foundation value types as fields.** `Date`, `URL`, `UUID`, `Data`, `Decimal` raise a compile-time `unknownNonPrimitiveType` error in v1.0+. See the *Foundation types* section below for the two supported workarounds.
 - [ ] **No generics with unbounded type parameters.** The macro reads syntactic types and won't infer constraints; if you need `MyType<T>`, give it `<T: Decodable & Sendable>` and don't expect the macro to do that for you.
@@ -206,7 +206,7 @@ Other knobs:
 
 The `@Castable` macro recognizes a fixed set of primitives (`String`, `Int`, `Double`, `Float`, `Bool`, arrays of those, plus other `@Castable` structs). Foundation value types — `Date`, `URL`, `UUID`, `Data`, `Decimal` — are **not** primitives from the macro's perspective: it can't synthesize a JSON Schema or grammar for them.
 
-> **Breaking change in v1.0:** prior versions emitted a *warning* and let expansion proceed; the consumer's downstream compile error was `<TypeName>.PartiallyGenerated.PartiallyGenerated?` — cryptic and hard to trace. As of v1.0, an `unknownNonPrimitiveType` field is a compile **error** at the macro site so the diagnostic is what you see, not a chain reaction in synthesized code.
+> **Breaking change in the v1.0 line (currently on `stage`, not yet released):** prior versions emitted a *warning* and let expansion proceed; the consumer's downstream compile error was `<TypeName>.PartiallyGenerated?` — cryptic and hard to trace. As of v1.0, an `unknownNonPrimitiveType` field is a compile **error** at the macro site so the diagnostic is what you see, not a chain reaction in synthesized code.
 
 ### Workaround 1: pre-convert at the model boundary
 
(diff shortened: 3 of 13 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,44 @@
...
+## [Unreleased]
+
+## [0.1.0] - 2026-04-30
+
+First public release. Cast ships as an Apache-2.0 Swift Package: type-safe structured output from any local LLM on Apple Silicon via constrained decoding.
+
+### Added
+
+- `@Castable` macro that synthesizes JSON Schema, `Decodable`, and a `PartiallyGenerated` mirror from an annotated struct.
+- `CastModel` with `cast`, `castJSON`, `classify`, `extract`, and `castStream` entrypoints; `init(wrapping:configuration:)` for caller-managed `ModelContainer` lifetimes.
```
jaylann: Factual drift against `git log v0.1.0`: at tag `v0.1.0`, `Sources/Cast/API/` had no `CastModel+Extract.swift` and no `CastModel+Stream.swift`. Both `extract` (issue #36, landed in `b061d53`) and `castStream` (issue #35, landed in `73df953`) shipped on `stage` *after* the v0.1.0 tag. Verified: `git ls-tree -r v0.1.0 -- Sources/Cast/API/ | rg -i 'extract|stream'` returns nothing, and `git show v0.1.0:Sources/Cast/API/CastModel+Generation.swift | rg 'func '` only shows `cast`/`castJSON`/`classify`.

Either move `extract` and `castStream` into the `[Unreleased]` section, or rephrase this bullet to match what actually shipped in 0.1.0 (`cast`, `castJSON`, `classify`, plus `init(wrapping:configuration:)`). Same goes for the related claims downstream — keep `[Unreleased]` honest so the next tag has somewhere to record these as features rather than backfilling them onto an already-published release.

CHANGELOG.md
```diff
@@ -0,0 +1,44 @@
...
+- Property wrappers covering schema constraints: `@MaxLength`/`@MinLength`, `@Count`/`@MaxCount`/`@MinCount`, `@CastRange`, `@Precision`, `@Pattern`, `@OneOf`, `@Description`, `@Examples`, `@Nullable`, `@DefaultValue`, `@Validator`.
+- `JSONSchema.excluding(fields:)` for dynamic schema modification at call time.
+- `CastModel.prepare(_:)` warm-up to pay grammar-compilation cost upfront.
+- `didGenerate` callback for cooperative cancellation; `abortInFlight()` plus iOS background-safety hooks (`enableBackgroundSafety()` / `disableBackgroundSafety()`).
+- DocC documentation site auto-generated from runnable `Examples/` snippets (9 example targets).
```
jaylann: `9 example targets` is off — `git show v0.1.0:Examples/Package.swift | rg 'executableTarget'` returns 11, and `git ls-tree -r v0.1.0 -- Sources/Cast/Cast.docc/Examples/` lists 11 articles. Either drop the parenthetical (it's brittle anyway — the count drifts every time someone adds an example) or make it `(11 example targets at release)`.

MIGRATION.md
```diff
@@ -21,8 +21,9 @@ For each existing type you want to migrate, walk these in order:
 
 - [ ] **Is it a `struct`?** Classes, actors, and `enum` (other than raw-value via `CastEnum`) are not supported by `@Castable`. The macro emits a `requiresStruct` diagnostic for non-structs.
 - [ ] **Are all stored properties `Sendable`?** Closures, `class` references, `weak` refs, `NSObject` subclasses, and most `AnyObject`-typed values are not.
-- [ ] **Are all stored properties `Decodable`?** `URL`, primitives, arrays of `Decodable`, optionals of `Decodable`, and other `@Castable` types are fine. Custom types with `init(from:)` are fine.
+- [ ] **Are all stored properties `Decodable`?** Primitives (`String`, `Int`, `Double`, `Float`, `Bool`), arrays of `Decodable`, optionals of `Decodable`, and other `@Castable` types are fine. Custom types with `init(from:)` are fine.
```
jaylann: This bullet still says "Custom types with `init(from:)` are fine" but the very next checklist item (line 26, added in this PR) tells the reader that `Date`/`URL`/`UUID`/`Data`/`Decimal` — all of which are `Decodable` via `init(from:)` — now compile-error. A reader walking the list top-to-bottom will check this box for `Date`, then uncheck it on the next line. Worth tightening the qualifier here (e.g. "Custom *user-defined* types with `init(from:)` are fine; Foundation value types are covered separately below") so the two bullets don't read as contradictory.

MIGRATION.md
```diff
@@ -201,6 +202,50 @@ Other knobs:
 
 ---
 
+## 3. Foundation types
+
+The `@Castable` macro recognizes a fixed set of primitives (`String`, `Int`, `Double`, `Float`, `Bool`, arrays of those, plus other `@Castable` structs). Foundation value types — `Date`, `URL`, `UUID`, `Data`, `Decimal` — are **not** primitives from the macro's perspective: it can't synthesize a JSON Schema or grammar for them.
+
+> **Breaking change in v1.0:** prior versions emitted a *warning* and let expansion proceed; the consumer's downstream compile error was `<TypeName>.PartiallyGenerated.PartiallyGenerated?` — cryptic and hard to trace. As of v1.0, an `unknownNonPrimitiveType` field is a compile **error** at the macro site so the diagnostic is what you see, not a chain reaction in synthesized code.
```
jaylann: Two small things on this callout:

1. "Breaking change in v1.0" — worth a phrase clarifying that this is a v1.0 *pre-release* breakage relative to 0.1.0 consumers, not retroactive. Anyone reading the published v0.1.0 docs would otherwise wonder if their existing code suddenly stopped compiling. Something like "Breaking change in the v1.0 line (currently on `stage`, not yet released)."
2. The error chain is described as `<TypeName>.PartiallyGenerated.PartiallyGenerated?` (double `.PartiallyGenerated`), but the diagnostic message in `CastableDiagnostic.swift` and the prior synthesized output in the macro test use the single-`.PartiallyGenerated?` form. One of the two is the real shape — pick one and be consistent across the migration note + the diagnostic text.

Sources/CastMacros/CastableDiagnostic.swift
```diff
@@ -55,11 +56,6 @@ enum CastableDiagnostic: DiagnosticMessage {
...
 
     var severity: DiagnosticSeverity {
-        switch self {
-        case .unknownNonPrimitiveType:
-            .warning
-        default:
-            .error
-        }
+        .error
     }
```
jaylann, line 68: Now that every case is `.error`, `severity` collapses to a single literal — fine for today, but the corresponding learning in CLAUDE.md (around line 131: the "`hasError = diagnostics.contains { $0.0.severity == .error }`" pattern in `CastableMacro.expansion`) is now load-bearing for a hypothetical future warning case that *can't* exist with this shape. Two options worth picking between explicitly:

- Keep the `switch self` form (with one branch, defaulting to `.error`) so the next contributor who adds a warning-grade diagnostic doesn't accidentally promote it to an error by inheriting the literal.
- Or document on the property that the contract is intentionally "every diagnostic is fatal," and update the CLAUDE.md learning to reflect that the per-case logic was removed.

Either is fine; the silent doc-rot is the actual concern.

Tests/CastMacroTests/CastDiagnosticTests.swift
```diff
@@ -181,28 +181,10 @@ struct CastDiagnosticTests {
...
-                init() {
-                }
-
-                struct PartiallyGenerated: Sendable, Decodable {
-                    var when: Date.PartiallyGenerated?
-                }
-            }
-
-            extension Event: Castable, Decodable {
             }
```
jaylann, line 184: Worth a sanity check that `assertMacroExpansion` from swift-syntax actually accepts `expandedSource` = the input source verbatim when the macro early-returns `[]`. The two plausible expected shapes are (a) input source unchanged (what's here) or (b) `expandedSource: ""`. swift-syntax's behavior is that an attached macro that returns no members produces no synthesized text, leaving the original declaration intact — so (a) should be correct, but please confirm this test actually runs green locally (`swift test --filter CastDiagnosticTests/unknownFoundationType`) before merging. If `assertMacroExpansion` instead expects an empty string here, the test will silently mis-match and we'd lose coverage of the early-return path.

Tests/CastTests/SchemaExcludingTests.swift
```diff
@@ -165,14 +170,187 @@ struct SchemaExcludingTests {
...
+                ("a", inner),
+                ("b", JSONSchema.array(items: inner)),
+                ("c", JSONSchema.string())
+            ),
+            required: ["a", "b", "c"],
+            additionalProperties: .boolean(false)
+        )
+
+        let reduced = schema.excluding(fields: ["c"])
+        let dict = try schemaDict(reduced)
```
jaylann, line 301: The test name promises a regression lock against `__N__` deduplication markers, but the assertions can't actually observe markers: `schemaDict(_:)` (line 348) regex-strips `__[0-9]+__` from the encoded JSON before calling `JSONSerialization.jsonObject(...)`. So this test passes whether `JSONSchema.excluding(fields:)` strips markers or not — the helper does the work either way. Effectively it's just another "top-level field exclusion preserves siblings" test on a schema that *would* have markers if the test inspected raw bytes.

Two ways to make it actually lock the marker-handling invariant:
1. Use a separate raw-bytes helper here that does *not* strip markers, and assert the parsed dict contains no `__N__`-bearing keys (or that `JSONSerialization` succeeds at all on the unsanitized output).
2. Or rename to `excluding handles schemas with structurally-shared sub-schemas` and drop the marker-specific framing — the test would still be a useful regression for sibling reuse, just not for the marker stripping.

Tests/CastTests/SchemaExcludingTests.swift
```diff
@@ -165,14 +170,187 @@ struct SchemaExcludingTests {
...
+    //
+    // These tests exist to fail loudly if the underlying JSON-string round-trip
+    // in `JSONSchema.excluding(fields:)` ever drifts (upstream `JSONSchema`
+    // bumps, encoder shape changes, regex-strip semantics). They lock structural
+    // properties — nested traversal, key adjacency, idempotence — that the
+    // behavioral tests above don't cover.
+
+    @Test("excluding preserves nested object property schemas")
+    func preservesNestedObject() throws {
+        let address = JSONSchema.object(
```
jaylann, line 184: Minor: the assertions here check that the *nested* `address.required` array still contains `street` and `zip`, but don't assert that the *top-level* `required` array correctly drops `name` while keeping `address`. That's the shape that would fail if the bug-class "excluding forgets to recurse into properties but still updates required" ever regressed in the opposite direction (forgets to update required, recurses correctly). Probably worth one extra `#expect` on `dict["required"]` to fully lock the surface.

Tests/CastTests/SchemaExcludingTests.swift
```diff
@@ -165,14 +170,187 @@ struct SchemaExcludingTests {
...
+
+        let once = schema.excluding(fields: ["b"])
+        let twice = once.excluding(fields: ["b"])
+
+        let onceProps = try (schemaDict(once))["properties"] as? [String: Any] ?? [:]
+        let twiceProps = try (schemaDict(twice))["properties"] as? [String: Any] ?? [:]
+        let onceRequired = try Set(schemaDict(once)["required"] as? [String] ?? [])
+        let twiceRequired = try Set(schemaDict(twice)["required"] as? [String] ?? [])
+
+        #expect(Set(onceProps.keys) == Set(twiceProps.keys))
```
jaylann, line 333: The comment on line 304 explains why byte-equality was abandoned, which is good context. One more property worth locking here while the test is fresh: assert the *required* arrays are equal as sets across `once` / `twice` (`onceRequired == twiceRequired` is already there — good) AND that they don't contain `b`. Right now if `excluding` regressed to leaving the excluded key in `required` while removing it from `properties`, both `once` and `twice` would carry the same broken `required`, and the property-set assertion would still pass. A `#expect(!onceRequired.contains("b"))` shores that up.

jaylann (commented):
Drive-by review of the cleanup batch. A couple of factual issues in the 0.1.0 backfill (CHANGELOG attributes post-release APIs to v0.1.0) plus one regression test whose helper short-circuits the very behavior the test name claims to lock. Rest is fine. Nothing blocking — leaving disposition to the maintainer.

Sources/CastMacros/CastableDiagnostic.swift
```diff
@@ -55,11 +56,6 @@ enum CastableDiagnostic: DiagnosticMessage {
...
 
     var severity: DiagnosticSeverity {
-        switch self {
-        case .unknownNonPrimitiveType:
-            .warning
-        default:
-            .error
-        }
+        .error
```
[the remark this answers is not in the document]
jaylann: Intentional behavior from #100 — the heuristic-vs-strict tradeoff was the explicit design decision. Leaving open for user judgment on whether to soften back.

Sources/CastMacros/CastableDiagnostic.swift
```diff
@@ -35,7 +36,7 @@ enum CastableDiagnostic: DiagnosticMessage {
         case .precisionOnNonFloat:
             "@Precision can only be applied to Double or Float properties"
         case let .unknownNonPrimitiveType(typeName):
-            "'\(typeName)' is not a Cast-supported primitive; the synthesized PartiallyGenerated mirror will project this field as '\(typeName).PartiallyGenerated?' and will fail to compile unless '\(typeName)' is itself @Castable. Consider wrapping it in a small @Castable struct or pre-converting to a primitive (e.g. ISO-8601 String for dates)."
+            "'\(typeName)' is not a Cast-supported type. Wrap it in a small @Castable struct, or pre-convert to a primitive at the model boundary (e.g. ISO-8601 String for Date, raw bytes for Data). See the 'Foundation types' section of MIGRATION.md."
```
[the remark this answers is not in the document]
jaylann: Fixed in 5a4a89e: changed Data suggestion to 'Base64 String or `[UInt8]`/`[Int]`' (both round-trip through JSON).

CONTRIBUTING.md
```diff
@@ -55,6 +55,21 @@ Use descriptive branch names:
...
+| File | Owns | Don't put here |
+|---|---|---|
+| `CastModel+Generation.swift` | every blocking generation entrypoint: `cast`, `castJSON`, `classify` | streaming, extraction-flavored prompts |
+| `CastModel+Stream.swift` | `castStream` and partial-decode helpers | blocking generation |
+| `CastModel+Extract.swift` | `extract(from:as:instruction:…)` and future extraction-shaped APIs | unrelated prompt templates |
+| `CastModel+Lifecycle.swift` | `abortInFlight()`, iOS background-safety hooks | model load/unload (the core `CastModel` type) |
+| `CastModel+Timeout.swift` | internal cross-isolation glue (`withGenerationTimeout`, `withInFlightRegistration`) | new public API |
+| `CastModel+GPUSafety.swift` | Metal/MLX lifecycle plumbing (`cleanupGPU`, global error handler) | code that doesn't touch `MLX.Stream` / `Memory` |
+
+If a new method doesn't fit any existing surface, add a new `CastModel+<Surface>.swift` rather than overloading one of the above. The first line of every such file should be a 2–3 line `// File rationale: …` comment that states what it owns and what it doesn't.
```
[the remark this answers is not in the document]
jaylann: Fixed in 5a4a89e: tightened CONTRIBUTING.md guidance to 'short header'.

CHANGELOG.md
jaylann: Fixed in 5a4a89e: verified via `git show v0.1.0:Sources/Cast/API/CastModel+Generation.swift | rg 'func '` — at v0.1.0 only `cast`/`castJSON`/`classify` shipped. Moved `extract` and `castStream` into [Unreleased].

CHANGELOG.md
jaylann: Fixed in 5a4a89e: confirmed 11 (not 9) executableTargets at v0.1.0; dropped the parenthetical entirely so the count can't drift again.

MIGRATION.md
jaylann: Fixed in 5a4a89e: scoped the bullet to 'user-defined types with init(from:)' and explicitly defer Foundation value types to the section below.

MIGRATION.md
jaylann: Fixed in 5a4a89e: clarified to 'Breaking change in the v1.0 line (currently on `stage`, not yet released)' and corrected the typo to `<TypeName>.PartiallyGenerated?`.

Sources/CastMacros/CastableDiagnostic.swift
jaylann: Fixed in 5a4a89e: restored `switch self { default: .error }` form with a comment pointing future contributors at the matching `hasError` early-return guard.

Tests/CastMacroTests/CastDiagnosticTests.swift
jaylann: Verified empirically — `CI=true swift test --filter CastMacroTests` green (26/26 tests, including the Foundation Date diagnostic test). `assertMacroExpansion` does accept the input source verbatim as expected when the macro early-returns `[]`.

Tests/CastTests/SchemaExcludingTests.swift
jaylann: Fixed in 5a4a89e: renamed to 'top-level field exclusion preserves siblings on schemas with shared sub-schemas' and added a comment noting that `schemaDict(_:)` regex-strips markers, so this test does not lock marker-handling inside `excluding` itself.

Tests/CastTests/SchemaExcludingTests.swift
jaylann: Fixed in 5a4a89e: added `#expect(\!topRequired.contains("name"))` and `#expect(topRequired.contains("address"))` to lock top-level required handling alongside the nested assertions.

Tests/CastTests/SchemaExcludingTests.swift
jaylann: Fixed in 5a4a89e: added `#expect(\!onceRequired.contains("b"))` so a regression that left excluded keys in `required` would fail this test.

PR #110, opened 2026-05-01
test(api): cover GrammarProcessorCache, castStream, Validator integration (#91, #93, #94)

### LINKED ISSUE #91: test(api): GrammarProcessorCache concurrent-load and cache-hit unit tests
`Tests/CastTests/CacheTests.swift` is 21 lines and exercises only `clear()` (smoke) and `prepareNoModel()` (error case). The interesting behavior of `Sources/Cast/API/GrammarProcessorCache.swift` is untested:

- Cached hits: a second `artifacts(for:)` call for a known key returns the cached value without re-loading.
- In-flight de-duplication: concurrent calls for the same key share one `Task`.
- Failure path: if the inner `Task` throws, the entry is removed from `inFlight` so the next call retries.
- `warmUp(for:)` populates the cache.

## Suggested fix

Refactor the loader behind a tiny protocol so a fake artifacts-loader can be injected at the actor boundary — no Metal, no real model download. Then write Swift Testing cases for each of the four behaviors above.

```swift
protocol ArtifactsLoading: Sendable {
    func load(_ config: ModelConfiguration) async throws -> TokenizerArtifacts
}
```

`GrammarMaskedLogitProcessor.loadTokenizerArtifacts` becomes the production conformer; tests inject a fake.

## Verification

- New tests run on CI (no `.requiresMetal`) and pass.
- Coverage of `GrammarProcessorCache.swift` jumps from ~15% to ~100%.

### LINKED ISSUE #93: test(api): non-Metal coverage of castStream incremental decoder
`Tests/CastTests/CastStreamTests.swift` has exactly one test — and it's `.requiresMetal`-gated, so CI never runs any streaming coverage at all.

## What's untested

The non-Metal half of `Sources/Cast/API/CastModel+Stream.swift` is meaty and worth direct tests:

- The incremental JSON-repair-and-decode loop (which bytes turn into the first valid `PartialResult`?).
- `BufferingPolicy.bufferingNewest(1)` semantics under fast producer / slow consumer.
- Cancellation surfaces `CastError.cancelled(partialOutput:)` carrying the most recent decoded snapshot.
- Timeout surfaces `CastError.timedOut(partialOutput:)` similarly.
- `progress` is monotonic non-decreasing.

## Suggested fix

`StreamBufferHolder` (line 188, currently `private final class`) is the seam — promote it (or its decode helper) to internal so `@testable import Cast` can drive it with a synthetic byte stream. No model, no Metal, no network.

## Verification

- New non-Metal test cases run on CI and pass.
- The existing `.requiresMetal` end-to-end test stays as-is for local Metal runs.

### LINKED ISSUE #94: test(api): integration test that Validator transformer fires during cast()
`Tests/CastTests/ValidatorTests.swift` covers `ValidatorSupport.decode(...)` directly and verifies the `@Validator` wrapper round-trips through `Codable`. **No test calls `model.cast(...)` on a `@Castable` containing `@Validator`** to confirm the transformer is wired into the live decode path.

This is a hidden integration gap: a refactor that drops the `ValidatorSupport.decode(...)` call from `cast()`'s decode flow would not fail any existing test, even though every `@Validator` would silently stop firing.

## Suggested fix

Add a unit test that decodes a fixed JSON string through the same code path `cast()` uses post-generation. No Metal needed — feed the raw JSON directly into the decode helper.

```swift
@Test("Validator runs during the cast() decode path")
func validatorIntegration() async throws {
    @Castable struct Score {
        @Validator({ max(0, min(100, $0)) }) var v: Int = 0
    }
    let raw = "{\"v\": 250}"
    // call into the actual decode helper used by cast(),
    // not ValidatorSupport.decode directly
    let s: Score = try decodeViaCastPath(raw)
    #expect(s.v == 100)
}
```

## Verification

- Test fails before the fix (if you stub it on a temporary branch that bypasses ValidatorSupport in cast()).
- Test passes against current `stage` HEAD.

jaylann:
## Summary

Three closely related test/refactor improvements bundled in one PR.

### #91 — GrammarProcessorCache concurrent-load + cache-hit unit tests
- New `TokenizerArtifactsLoader` closure seam; default points at `GrammarMaskedLogitProcessor.loadTokenizerArtifacts(configuration:)`.
- New tests cover cache hit, in-flight de-duplication, failure-path inflight cleanup, `warmUp`, and `clear`.
- Adds a public memberwise init to vendored `TokenizerArtifacts` (`// Modifications:` marker per the vendoring rule).

### #93 — non-Metal coverage of castStream incremental decoder
- Relaxes visibility on `StreamBufferHolder`, `decodePartial`, `yieldTerminal` (private → internal) so non-Metal tests can drive them.
- New tests cover repair-and-decode, monotonic progress, progress clamp at 1.0, `StreamBufferHolder` concurrency, terminal-yield repair behavior, decode-failure paths, and `AsyncThrowingStream` `bufferingNewest(1)` semantics.
- The existing `.requiresMetal` end-to-end test stays unchanged.

### #94 — Validator integration test through cast() decode path
- Extracts `CastDecode.decode<T: Decodable>(_:rawOutput:config:)` from `cast()`'s post-generation block.
- Pure refactor; same control flow.
- New test directly exercises the helper through a `Castable` struct with `@Validator`, proving the transformer fires through the same code path `cast()` runs.

## Test plan

- [x] `CI=true swift test` passes (218 tests, 16 suites, 0 failures locally).
- [x] No new test uses `.requiresMetal` — all run on CI.

Closes #91
Closes #93
Closes #94

### DIFF
--- a/Sources/Cast/API/CastModel+Generation.swift
+++ b/Sources/Cast/API/CastModel+Generation.swift
@@ -318,25 +318,24 @@ public extension CastModel {
 /// tests can drive the same repair + ValidatorSupport.decode path that
 /// production generation uses, without spinning up MLX.
 enum CastDecode {
+    static func repair(rawOutput: String, config: CastConfiguration) throws -> String {
+        guard config.repairTruncatedJSON else { return rawOutput }
+        switch JSONRepair.repair(rawOutput) {
+        case let .ok(value):
+            return value
+        case let .repaired(value, _):
+            return value
+        case let .unrecoverable(reason):
+            throw CastError.repairFailed(rawOutput: rawOutput, reason: reason)
+        }
+    }
+
     static func decode<T: Decodable>(
         _: T.Type,
         rawOutput: String,
         config: CastConfiguration
     ) throws -> T {
-        let decodeInput: String
-        if config.repairTruncatedJSON {
-            switch JSONRepair.repair(rawOutput) {
-            case let .ok(value):
-                decodeInput = value
-            case let .repaired(value, _):
-                decodeInput = value
-            case let .unrecoverable(reason):
-                throw CastError.repairFailed(rawOutput: rawOutput, reason: reason)
-            }
-        } else {
-            decodeInput = rawOutput
-        }
-
+        let decodeInput = try repair(rawOutput: rawOutput, config: config)
         do {
             return try ValidatorSupport.decode(T.self, from: Data(decodeInput.utf8))
         } catch {
--- a/Sources/Cast/API/CastModel+Generation.swift
+++ b/Sources/Cast/API/CastModel+Generation.swift
@@ -248,28 +248,7 @@ public extension CastModel {
             throw CastError.cancelled(partialOutput: result.output)
         }
 
-        let decodeInput: String
-        if config.repairTruncatedJSON {
-            switch JSONRepair.repair(result.output) {
-            case let .ok(value):
-                decodeInput = value
-            case let .repaired(value, _):
-                decodeInput = value
-            case let .unrecoverable(reason):
-                throw CastError.repairFailed(rawOutput: result.output, reason: reason)
-            }
-        } else {
-            decodeInput = result.output
-        }
-
-        do {
-            return try ValidatorSupport.decode(T.self, from: Data(decodeInput.utf8))
-        } catch {
-            throw CastError.decodingFailed(
-                rawOutput: decodeInput,
-                error: error.localizedDescription
-            )
-        }
+        return try CastDecode.decode(T.self, rawOutput: result.output, config: config)
     }
 
     /// Classify `prompt` into one of the cases of a ``CastEnum`` with a
@@ -332,5 +311,39 @@ public extension CastModel {
             built.user, as: type, schema: schema, system: built.system,
             config: classifyConfig, didGenerate: didGenerate
         )
+    }
+}
+
+/// Post-generation decode helper. Extracted from `cast(_:as:schema:...)` so
+/// tests can drive the same repair + ValidatorSupport.decode path that
+/// production generation uses, without spinning up MLX.
+enum CastDecode {
+    static func decode<T: Decodable>(
+        _: T.Type,
+        rawOutput: String,
+        config: CastConfiguration
+    ) throws -> T {
+        let decodeInput: String
+        if config.repairTruncatedJSON {
+            switch JSONRepair.repair(rawOutput) {
+            case let .ok(value):
+                decodeInput = value
+            case let .repaired(value, _):
+                decodeInput = value
+            case let .unrecoverable(reason):
+                throw CastError.repairFailed(rawOutput: rawOutput, reason: reason)
+            }
+        } else {
+            decodeInput = rawOutput
+        }
+
+        do {
+            return try ValidatorSupport.decode(T.self, from: Data(decodeInput.utf8))
+        } catch {
+            throw CastError.decodingFailed(
+                rawOutput: decodeInput,
+                error: error.localizedDescription
+            )
+        }
     }
 }
--- a/Sources/Cast/API/CastModel+Stream.swift
+++ b/Sources/Cast/API/CastModel+Stream.swift
@@ -103,7 +103,7 @@ public extension CastModel {
 
         // Cross-isolation handle for the in-flight buffer so the cancellation
         // catch (outside the perform closure) can recover the partial bytes.
-        let bufferHolder = StreamBufferHolder()
+        let bufferHolder = StreamDecode.BufferHolder()
         do {
             try await withInFlightRegistration {
                 try await withGenerationTimeout(config.timeout) {
@@ -128,7 +128,7 @@ public extension CastModel {
                             lastTokenCount = chunk.totalTokens
                             bufferHolder.set(buffer)
 
-                            if let snapshot = decodePartial(
+                            if let snapshot = StreamDecode.partial(
                                 T.self,
                                 from: buffer,
                                 tokenCount: lastTokenCount,
@@ -146,7 +146,7 @@ public extension CastModel {
                         // outer `Task.isCancelled` check below instead.
                         if Task.isCancelled { return }
 
-                        try yieldTerminal(
+                        try StreamDecode.terminal(
                             T.self,
                             buffer: buffer,
                             tokenCount: lastTokenCount,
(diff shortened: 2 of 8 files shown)


Sources/Cast/API/GrammarProcessorCache.swift
```diff
@@ -17,8 +27,9 @@ actor GrammarProcessorCache {
             return try await existing.value
         }
 
+        let loader = loader
```
jaylann: Minor: the `let loader = loader` rebind is redundant here. `loader` is a stored `let` on the actor and is `@Sendable`, so the `Task { ... }` initializer can capture `self.loader` directly under Swift 6 isolation. Harmless, but worth a one-line comment if you want to keep the defensive shadow ("explicit local capture so Task does not retain self") so a future reader does not delete it as dead code.

Sources/Cast/API/CastModel+Stream.swift
```diff
@@ -202,7 +202,7 @@ private final class StreamBufferHolder: @unchecked Sendable {
     }
 }
 
-private func decodePartial<T: Castable>(
+func decodePartial<T: Castable>(
```
jaylann: Visibility relaxation is the minimum that `@testable import Cast` can reach (private symbols are not exposed even under `@testable`), so this is correct. However `decodePartial` and `yieldTerminal` are now module-internal *free functions* — any other file in `Cast` can now call them by their bare name. Consider namespacing under a caseless enum (`enum StreamDecode { static func partial(...) ; static func terminal(...) }`) to keep the seam open for tests without polluting module-internal namespace.

Sources/Cast/API/CastModel+Generation.swift
```diff
@@ -248,28 +248,7 @@ public extension CastModel {
...
-
-        do {
-            return try ValidatorSupport.decode(T.self, from: Data(decodeInput.utf8))
-        } catch {
-            throw CastError.decodingFailed(
-                rawOutput: decodeInput,
-                error: error.localizedDescription
-            )
-        }
+        return try CastDecode.decode(T.self, rawOutput: result.output, config: config)
```
jaylann, line 251: Drive-by: `CastModel+Stream.swift` `yieldTerminal` (lines 243–272) duplicates the exact `repairTruncatedJSON` switch + repair-failure mapping that `CastDecode.decode` now owns. Followup: have `yieldTerminal` reuse `CastDecode.decode(T.self, rawOutput: buffer, config: config)` to enforce `T`-required-field semantics, then re-decode `T.PartiallyGenerated` from the repaired bytes. Removes ~30 lines of drift risk between cast() and castStream() terminal yields. Worth filing.

Tests/CastTests/CastStreamTests.swift
```diff
@@ -35,3 +35,206 @@ func castStreamProducesPartialsAndTerminal() async throws {
...
+                StreamMovie.self,
+                buffer: "{\"title\": \"Inception\", \"year\": 2010",
+                tokenCount: 50,
+                maxTokens: 100,
+                minProgress: 0,
+                config: config,
+                continuation: streamContinuation
+            )
+        }
+    }
```
jaylann, line 196: The test name says "throws decodingFailed when repair is off" but the assertion is the generic `CastError.self`. Tighten with the same case-pattern matcher you use in `yieldTerminalThrowsRepairFailedOnUnrecoverable` (lines 211–218): assert `case .decodingFailed = castError` so a future bug that maps this path to `repairFailed` or `generationFailed` does not slip through.

Tests/CastTests/CastStreamTests.swift
```diff
@@ -35,3 +35,206 @@ func castStreamProducesPartialsAndTerminal() async throws {
...
+        #expect(collected.count == 1)
+        let only = try #require(collected.first)
+        #expect(only.value.title == "Inception")
+        #expect(only.value.year == 2010)
+    }
+
+    @Test("yieldTerminal throws decodingFailed when repair is off")
+    func yieldTerminalThrowsOnRepairFailure() {
+        var streamContinuation: AsyncThrowingStream<PartialResult<StreamMovie>, Error>.Continuation!
+        _ = AsyncThrowingStream<PartialResult<StreamMovie>, Error> { continuation in
```
jaylann: `_ = AsyncThrowingStream<...> { ... }` discards the stream and only retains the captured `streamContinuation`. Works because `yieldTerminal` only throws (no yield) on this path, so the dropped stream + dangling continuation never surface a problem. Same pattern at line 194. If you ever change `yieldTerminal` to flush partials before throwing, these two tests will silently lose those yields. Either bind the stream to a local `let` (matching the success-path test at line 140) or add an inline comment explaining the discard is intentional.

Tests/CastTests/CacheTests.swift
```diff
@@ -13,9 +21,95 @@ struct CacheTests {
...
+        #expect(count == 1)
+        #expect(results.0.vocab == results.1.vocab)
+    }
+
+    @Test("failure path clears in-flight so next call retries")
+    func failurePathClearsInflight() async throws {
+        let counter = LoaderCounter()
+        let cache = GrammarProcessorCache { _ in
+            await counter.bump()
+            let invocation = await counter.count
```
jaylann: Subtle: `let invocation = await counter.count` after `await counter.bump()` is two separate actor hops. Today this is sequential because the test calls `cache.artifacts(...)` serially, but if you ever extend this test to drive concurrent failure behaviour, two loader invocations could both observe `count == 1` (or both observe `2`) between the bump and the read, racing on whether to throw. Safer pattern: have `bump()` *return* the incremented count atomically (`func bump() -> Int { count += 1; return count }`).

jaylann (commented):
Solid PR — three independent testability improvements landing cleanly. Loader-injection seam is well-designed (default closure points at the production loader; tests can sub a counter). CastDecode extraction preserves error mapping verbatim. Vendored upstream change is minimal and properly attributed. A few minor observations inline; none blocking.

Sources/Cast/API/GrammarProcessorCache.swift
jaylann: Fixed in d215373: dropped the `let loader = loader` rebind. The actor's `let loader: TokenizerArtifactsLoader` is `@Sendable`, so `Task { try await loader(configuration) }` captures it directly under Swift 6 isolation.

Sources/Cast/API/CastModel+Stream.swift
jaylann: Fixed in d215373: introduced `enum StreamDecode { static func partial(...); static func terminal(...); final class BufferHolder }` at file scope and moved the three internals into it. Module-internal namespace is no longer polluted with bare `decodePartial` / `yieldTerminal` / `StreamBufferHolder`.

Sources/Cast/API/CastModel+Generation.swift
jaylann: Done in d215373: factored `CastDecode.repair(rawOutput:config:) throws -> String` out of the existing `decode<T>(...)` (which now composes repair + ValidatorSupport.decode), then rewrote `StreamDecode.terminal` to use `CastDecode.repair` + `ValidatorSupport.decode(T.self, ...)` for required-field enforcement + `JSONDecoder().decode(T.PartiallyGenerated.self, ...)` for the consumer-facing type. Drift between `cast()` and `castStream()` terminal yields is eliminated.

Tests/CastTests/CastStreamTests.swift
jaylann: Fixed in d215373: tightened to the same `case .decodingFailed = castError` matcher pattern used at the `yieldTerminalThrowsRepairFailedOnUnrecoverable` test below. A regression that mapped this path to `.repairFailed` or `.generationFailed` will now fail the test.

Tests/CastTests/CastStreamTests.swift
jaylann: Fixed in d215373: bound the streams to `let _stream = AsyncThrowingStream<...> { ... }` (matching the success-path pattern at line 140) so a future change to `terminal` that flushes partials before throwing actually exercises the full stream rather than dropping it on the floor.

Tests/CastTests/CacheTests.swift
jaylann: Fixed in d215373: changed `LoaderCounter.bump()` to `func bump() -> Int { count += 1; return count }` and rewrote the failure-path closure as `let invocation = await counter.bump(); if invocation == 1 { throw ... }`. Atomic single hop.

Sources/Cast/API/GrammarProcessorCache.swift
```diff
@@ -2,9 +2,19 @@ import Foundation
 import MLXLMCommon
 @preconcurrency import MLXStructured
 
+typealias TokenizerArtifactsLoader =
```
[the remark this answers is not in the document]
jaylann: Fixed in d215373: prefixed `typealias TokenizerArtifactsLoader` with `private` at file scope. Side note — the actor's `init` had to be desugared from `init(loader: @escaping TokenizerArtifactsLoader = ...)` to `init(loader: @escaping @Sendable (ModelConfiguration) async throws -> TokenizerArtifacts = ...)` because Swift forbids an internal init from referencing a private typealias in its parameter type (the compiler errors with `initializer must be declared fileprivate because its parameter uses a private type`). The internal stored property still uses the typealias for readability.

Tests/CastTests/CastStreamTests.swift
```diff
@@ -35,3 +35,206 @@ func castStreamProducesPartialsAndTerminal() async throws {
...
+        }
+
+        var collected: [Int] = []
+        for try await value in stream {
+            collected.append(value)
+        }
+
+        #expect(collected.count == 1)
+        #expect(collected.first == 5)
+    }
```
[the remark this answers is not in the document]
jaylann: Fixed in d215373: dropped the test entirely — it was exercising stdlib `AsyncThrowingStream(bufferingNewest(1))` behavior in isolation, not `castStream`. The buffering invariant for `castStream` itself is exercised by the `castStreamProducesPartialsAndTerminal` Metal test.

Sources/Cast/API/CastModel+Stream.swift
```diff
@@ -231,7 +231,7 @@ private func decodePartial<T: Castable>(
     return PartialResult<T>(value: value, progress: progress, tokenCount: tokenCount)
 }
 
-private func yieldTerminal<T: Castable>(
+func yieldTerminal<T: Castable>(
     _: T.Type,
```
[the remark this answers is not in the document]
jaylann: Fixed in d215373: same as the thread on line 205 — `decodePartial`, `yieldTerminal`, and `StreamBufferHolder` are now nested under `enum StreamDecode { ... }`. No more bare module-internal generics.

PR #111, opened 2026-05-01
test(schema): direct unit tests for ZeroSchemaDecoder (#92)

### LINKED ISSUE #92: test(schema): direct unit tests for ZeroSchemaDecoder
`Sources/Cast/Schema/ZeroSchemaDecoder.swift` is 507 LoC and is the runtime backbone for converting a `Decodable & Sendable` type into a structural `SchemaInfo`. It has **no dedicated test file** — it's exercised only indirectly through `SchemaGenerator` happy-path tests.

## Coverage gaps

A dedicated `Tests/CastTests/ZeroSchemaDecoderTests.swift` should cover:

- Nested Castable types decoded into `SchemaKind`.
- Optional fields ending up out of the `required` list.
- Arrays (`[T]`), dictionaries (`[String: T]`).
- Enums via `_FirstCaseProvider` (currently public — verify the contract).
- Recursive types (struct that contains itself or a peer).
- Foundation types: `Date`, `URL`, `UUID`, `Data`, `Decimal` — confirm warning-path produces `<TypeName>.PartiallyGenerated?` projection per `CastableDiagnostic.unknownNonPrimitiveType`.
- `CastSchemaProviding` conformance — does a custom override get used?

## Verification

- New test file with coverage for each case above.
- All tests run without `.requiresMetal` (no model needed).
- Mutation testing or eye-check confirms decoder code paths are exercised.

jaylann:
## Summary

`Sources/Cast/Schema/ZeroSchemaDecoder.swift` (507 LoC) is the runtime backbone for converting a `Decodable` type into a structural `SchemaInfo`. It was previously exercised only indirectly through `SchemaGenerator` happy-path tests.

`Tests/CastTests/ZeroSchemaDecoderTests.swift` (new, 246 LoC) covers:

- All primitive types → matching `SchemaKind`
- Optional fields excluded from `required`
- Primitive arrays + arrays of nested structs
- Nested `@Castable` objects → `.object` kind with inner properties
- Enums via `CastEnum` + `_FirstCaseProvider` (zero instance returns first case)
- Property-wrapper short-circuit (`@MaxLength` etc. → wrapped value's kind, not `.object`)
- `CastSchemaProviding` override is honored over structural decoding
- Field order matches Decodable init traversal
- `_FirstCaseProvider` protocol surface
- `info.zeroInstance` populates primitives correctly
- Each `SchemaField.schema` round-trips through `JSONEncoder`

## Out of scope

Recursive types: Swift forbids value-type self/mutual reference; class-based recursion triggers infinite recursion in the zero decoder's `decodeIfPresent<T: Decodable>` path (no depth guard). Recursion is handled at macro time via `CastableDiagnostic.unknownNonPrimitiveType` instead.

Foundation types (Date/URL/UUID/Data/Decimal): macro-time concerns, not runtime decoder concerns.

## Test plan

- [x] `CI=true swift test` passes (212 tests, 16 suites, 0 failures locally).
- [x] No `.requiresMetal` traits used (pure runtime decoder logic).

Closes #92

### DIFF
--- a/Tests/CastTests/ZeroSchemaDecoderTests.swift
+++ b/Tests/CastTests/ZeroSchemaDecoderTests.swift
@@ -66,6 +66,11 @@ private struct WithMaxLen {
     @MaxLength(10) var title: String = ""
 }
 
+@Castable
+private struct WithRangeNumeric {
+    @CastRange(0...100) var n: Int = 0
+}
+
 private enum CustomKind: String, Decodable, Sendable, CaseIterable, CastSchemaProviding, _FirstCaseProvider {
     case alpha, beta
     static var castSchema: JSONSchema {
@@ -128,7 +133,9 @@ struct ZeroSchemaDecoderDirectTests {
     func requiredVsOptional() throws {
         let info = try ZeroSchemaDecoder.decode(RequiredAndOptional.self)
         #expect(info.required == ["a"])
-        #expect(info.fields.count == 2)
+        #expect(info.fields.map(\.name) == ["a", "b"])
+        let kinds = Dictionary(uniqueKeysWithValues: info.fields.map { ($0.name, $0.kind) })
+        #expect(kinds["b"] == .string)
     }
 
     @Test("Primitive arrays produce array schema with element kinds")
@@ -154,8 +161,7 @@ struct ZeroSchemaDecoderDirectTests {
         )
         let items = try #require(json["items"] as? [String: Any])
         let props = try #require(items["properties"] as? [String: Any])
-        let propKeys = props.keys.joined(separator: ",")
-        #expect(propKeys.contains("x"))
+        #expect(props.keys.contains(where: { $0.contains("x") }))
     }
 
     @Test("Nested @Castable object yields object kind with inner properties")
@@ -170,8 +176,7 @@ struct ZeroSchemaDecoderDirectTests {
         )
         #expect(json["type"] as? String == "object")
         let props = try #require(json["properties"] as? [String: Any])
-        let propKeys = props.keys.joined(separator: ",")
-        #expect(propKeys.contains("x"))
+        #expect(props.keys.contains(where: { $0.contains("x") }))
     }
 
     @Test("Enum via CastEnum decodes as enumeration with first case zero")
@@ -179,6 +184,11 @@ struct ZeroSchemaDecoderDirectTests {
         let info = try ZeroSchemaDecoder.decode(ColorHolder.self)
         let field = try #require(info.fields.first { $0.name == "color" })
         #expect(field.kind == .enumeration)
+
+        let data = try JSONEncoder().encode(field.schema)
+        let json = try #require(try JSONSerialization.jsonObject(with: data) as? [String: Any])
+        let enumValues = try #require(json["enum"] as? [String])
+        #expect(Set(enumValues) == ["red", "blue"])
 
         let zero = try #require(info.zeroInstance as? ColorHolder)
         #expect(zero.color == .red)
@@ -195,7 +205,13 @@ struct ZeroSchemaDecoderDirectTests {
         #expect(zero.f == 0)
         #expect(zero.i == 0)
         #expect(zero.i8 == 0)
+        #expect(zero.i16 == 0)
+        #expect(zero.i32 == 0)
+        #expect(zero.i64 == 0)
         #expect(zero.u == 0)
+        #expect(zero.u8 == 0)
+        #expect(zero.u16 == 0)
+        #expect(zero.u32 == 0)
         #expect(zero.u64 == 0)
     }
 
@@ -206,6 +222,13 @@ struct ZeroSchemaDecoderDirectTests {
         #expect(field.kind == .string)
     }
 
+    @Test("Property wrapper field on numeric short-circuits to integer kind")
+    func propertyWrapperNumericShortCircuit() throws {
+        let info = try ZeroSchemaDecoder.decode(WithRangeNumeric.self)
+        let field = try #require(info.fields.first { $0.name == "n" })
+        #expect(field.kind == .integer)
+    }
+
     @Test("CastSchemaProviding override is used over structural decoding")
     func castSchemaProvidingOverride() throws {
         let info = try ZeroSchemaDecoder.decode(CustomKindHolder.self)
@@ -213,9 +236,12 @@ struct ZeroSchemaDecoderDirectTests {
         #expect(field.kind == .enumeration)
 
         let data = try JSONEncoder().encode(field.schema)
-        let raw = String(decoding: data, as: UTF8.self)
-        #expect(raw.contains("ALPHA"))
-        #expect(raw.contains("BETA"))
+        let json = try #require(try JSONSerialization.jsonObject(with: data) as? [String: Any])
+        let enumValues = try #require(json["enum"] as? [String])
+        #expect(Set(enumValues) == ["ALPHA", "BETA"])
+
+        let zero = try #require(info.zeroInstance as? CustomKindHolder)
+        #expect(zero.kind == .alpha)
     }
 
     @Test("Field order matches Decodable init traversal, not alphabetical")
--- a/Tests/CastTests/ZeroSchemaDecoderTests.swift
+++ b/Tests/CastTests/ZeroSchemaDecoderTests.swift
@@ -1,0 +1,246 @@
+@testable import Cast
+import Collections
+import Foundation
+import JSONSchema
+import Testing
+
+// MARK: - Test Fixtures
+
+private struct AllPrimitives: Decodable, Sendable {
+    var b: Bool = false
+    var s: String = ""
+    var d: Double = 0
+    var f: Float = 0
+    var i: Int = 0
+    var i8: Int8 = 0
+    var i16: Int16 = 0
+    var i32: Int32 = 0
+    var i64: Int64 = 0
+    var u: UInt = 0
+    var u8: UInt8 = 0
+    var u16: UInt16 = 0
+    var u32: UInt32 = 0
+    var u64: UInt64 = 0
+}
+
+private struct RequiredAndOptional: Decodable, Sendable {
+    var a: Int = 0
+    var b: String?
+}
+
+private struct PrimitiveArrays: Decodable, Sendable {
+    var ss: [String] = []
+    var ii: [Int] = []
+    var dd: [Double] = []
+    var bb: [Bool] = []
+}
+
+private struct InnerArrayItem: Decodable, Sendable {
+    var x: Int = 0
+}
+
+private struct ArrayOfNested: Decodable, Sendable {
+    var items: [InnerArrayItem] = []
+}
+
+@Castable
+private struct InnerNest {
+    var x: Int = 0
+}
+
+@Castable
+private struct OuterNest {
+    var inner: InnerNest = .init()
+}
+
+private enum SwatchColor: String, CastEnum {
+    case red, blue
+}
+
+private struct ColorHolder: Decodable, Sendable {
+    var color: SwatchColor = .red
+}
+
+@Castable
+private struct WithMaxLen {
+    @MaxLength(10) var title: String = ""
+}
+
+private enum CustomKind: String, Decodable, Sendable, CaseIterable, CastSchemaProviding, _FirstCaseProvider {
+    case alpha, beta
+    static var castSchema: JSONSchema {
+        .enum(values: [.string("ALPHA"), .string("BETA")])
+    }
+
+    static var _firstCaseAny: Any {
+        CustomKind.alpha
+    }
+}
+
+private struct CustomKindHolder: Decodable, Sendable {
+    var kind: CustomKind = .alpha
+}
+
+private struct FieldOrder: Decodable, Sendable {
+    var z: Int = 0
+    var a: String = ""
+    var m: Bool = false
+}
+
+private enum Direction: String, Decodable, Sendable, CaseIterable, _FirstCaseProvider {
+    case north, south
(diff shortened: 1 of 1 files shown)


Tests/CastTests/ZeroSchemaDecoderTests.swift
```diff
@@ -0,0 +1,246 @@
...
+        #expect(kinds["u16"] == .integer)
+        #expect(kinds["u32"] == .integer)
+        #expect(kinds["u64"] == .integer)
+    }
+
+    @Test("Required vs optional tracked separately")
+    func requiredVsOptional() throws {
+        let info = try ZeroSchemaDecoder.decode(RequiredAndOptional.self)
+        #expect(info.required == ["a"])
+        #expect(info.fields.count == 2)
```
jaylann: This test asserts `required == ["a"]` and `fields.count == 2`, but never verifies that `b` is present in `info.fields` or that its kind is `.string`. A regression where `decodeIfPresent(String.Type, ...)` stopped appending to `fields` would still leave `count == 2` only if some other field were spuriously added — but the safer assertion is `#expect(info.fields.map(\.name) == ["a", "b"])` plus `#expect(kinds["b"] == .string)`. As written this only proves the `required` filter, not that optional fields round-trip into the schema.

Tests/CastTests/ZeroSchemaDecoderTests.swift
```diff
@@ -0,0 +1,246 @@
...
+
+    @Test("Enum via CastEnum decodes as enumeration with first case zero")
+    func enumViaCastEnum() throws {
+        let info = try ZeroSchemaDecoder.decode(ColorHolder.self)
+        let field = try #require(info.fields.first { $0.name == "color" })
+        #expect(field.kind == .enumeration)
+
+        let zero = try #require(info.zeroInstance as? ColorHolder)
+        #expect(zero.color == .red)
+    }
```
jaylann, line 195: Test name promises "decodes as enumeration with first case zero" but only checks `field.kind == .enumeration` and `zero.color == .red`. The schema content (that "red"/"blue" are emitted as the enum values) is never verified — a regression where `CastEnum.castSchema` stopped enumerating `allCases` would slip past. Either encode `field.schema` and assert the enum values, or rely on the `castSchemaProvidingOverride` test to cover that and rename this one.

Tests/CastTests/ZeroSchemaDecoderTests.swift
```diff
@@ -0,0 +1,246 @@
...
+    @Test("CastSchemaProviding override is used over structural decoding")
+    func castSchemaProvidingOverride() throws {
+        let info = try ZeroSchemaDecoder.decode(CustomKindHolder.self)
+        let field = try #require(info.fields.first { $0.name == "kind" })
+        #expect(field.kind == .enumeration)
+
+        let data = try JSONEncoder().encode(field.schema)
+        let raw = String(decoding: data, as: UTF8.self)
+        #expect(raw.contains("ALPHA"))
+        #expect(raw.contains("BETA"))
```
jaylann: Substring-matching the encoded JSON for "ALPHA"/"BETA" is brittle — it would silently miss schemas where the values land in an unrelated field, and gives no signal about the schema *shape* (still `enum`? still `string` items?). Prefer decoding to `[String: Any]` and asserting `json["enum"] as? [String] == ["ALPHA", "BETA"]`. The uppercase trick keeps it from colliding with the case names today, but the test would survive a regression where the override silently degraded to a plain string schema with `"ALPHA"` somewhere in it.

Tests/CastTests/ZeroSchemaDecoderTests.swift
```diff
@@ -0,0 +1,246 @@
...
+        #expect(zero.u == 0)
+        #expect(zero.u64 == 0)
+    }
+
+    @Test("Property wrapper field short-circuits to wrapped value's kind")
+    func propertyWrapperShortCircuit() throws {
+        let info = try ZeroSchemaDecoder.decode(WithMaxLen.self)
+        let field = try #require(info.fields.first { $0.name == "title" })
+        #expect(field.kind == .string)
+    }
```
jaylann, line 223: Single-wrapper coverage. The wrapper short-circuit in `schemaForWrapper` (decoder.swift:440) handles every `@MaxLength`/`@MinLength`/`@Range`/`@MaxCount`/`@MinCount`/`@Pattern`/etc. wrapper because they all expose `wrappedValue` via Mirror — but a test of one stringly-wrapped String mostly proves the String path. Worth adding a numeric case (e.g. `@Range(0...100) var n: Int`) to cover `schemaForValue` mapping non-string wrapped values, since the decoder distinguishes Int vs Double vs Bool there.

Tests/CastTests/ZeroSchemaDecoderTests.swift
```diff
@@ -0,0 +1,246 @@
...
+        let zero = try #require(info.zeroInstance as? AllPrimitives)
+
+        #expect(zero.b == false)
+        #expect(zero.s == "")
+        #expect(zero.d == 0)
+        #expect(zero.f == 0)
+        #expect(zero.i == 0)
+        #expect(zero.i8 == 0)
+        #expect(zero.u == 0)
+        #expect(zero.u64 == 0)
```
jaylann, line 215: Inconsistent with the kinds test above: `i16`, `i32`, `i64`, `u8`, `u16`, `u32` are checked for `kind == .integer` but their zero values are not asserted here. Either drop the unchecked variants from `AllPrimitives` or assert all 14 zero values for parity. As-is, a regression where `decode(UInt32.Type, ...)` returned `1` instead of `0` would be missed.

Tests/CastTests/ZeroSchemaDecoderTests.swift
```diff
@@ -0,0 +1,246 @@
...
+private enum CustomKind: String, Decodable, Sendable, CaseIterable, CastSchemaProviding, _FirstCaseProvider {
+    case alpha, beta
+    static var castSchema: JSONSchema {
+        .enum(values: [.string("ALPHA"), .string("BETA")])
+    }
+
+    static var _firstCaseAny: Any {
+        CustomKind.alpha
+    }
+}
```
jaylann, line 83: `CustomKind` conforms to `_FirstCaseProvider` directly (without going through `CastEnum`). This is good — it isolates the `CastSchemaProviding`-only-no-CastEnum path. But the corresponding test (`castSchemaProvidingOverride`, line 209) never asserts that `info.zeroInstance` materializes `CustomKindHolder` with `kind == .alpha`. The decoder explicitly calls `firstCase(of:)` on line 176 of `ZeroSchemaDecoder.swift`; that branch is only proven if you check the zero-instance value here.

jaylann (commented):
Solid coverage of the ZeroSchemaDecoder surface. A few places where the assertion is shallower than the test name implies — most notably the optional-field test doesn't actually verify the optional field is decoded, and the CastSchemaProviding override test relies on substring matching. Suggestions inline; non-blocking.

Tests/CastTests/ZeroSchemaDecoderTests.swift
jaylann: Fixed in e95823f: added `info.fields.map(\.name) == ["a", "b"]` and `kinds["b"] == .string`. Now genuinely proves `b` round-trips into the schema as a string field, not just that `required` filtered it out.

Tests/CastTests/ZeroSchemaDecoderTests.swift
jaylann: Fixed in e95823f: now encodes `field.schema`, parses to `[String: Any]`, and asserts `Set(json["enum"] as? [String]) == ["red", "blue"]`. A regression where `CastEnum.castSchema` stopped enumerating `allCases` will now fail.

Tests/CastTests/ZeroSchemaDecoderTests.swift
jaylann: Fixed in e95823f: replaced the substring match with a structured parse — decode `field.schema` to `[String: Any]` and assert `Set(json["enum"] as? [String]) == ["ALPHA", "BETA"]`. Schema shape is now what's checked, not whether "ALPHA" appears anywhere in the encoded bytes.

Tests/CastTests/ZeroSchemaDecoderTests.swift
jaylann: Fixed in e95823f: added a sibling test `propertyWrapperNumericShortCircuit` covering `@CastRange(0...100) var n: Int`, asserting `kind == .integer`. This exercises `schemaForValue`'s non-string branch which the String-only test doesn't reach.

Tests/CastTests/ZeroSchemaDecoderTests.swift
jaylann: Fixed in e95823f: filled in the missing 6 zero-value assertions (`i16`, `i32`, `i64`, `u8`, `u16`, `u32`). All 14 primitives now have parity between the kinds test and the zero-values test.

Tests/CastTests/ZeroSchemaDecoderTests.swift
jaylann: Fixed in e95823f: added `let zero = try #require(info.zeroInstance as? CustomKindHolder); #expect(zero.kind == .alpha)` to `castSchemaProvidingOverride`. The CastSchemaProviding-only-no-CastEnum branch (`firstCase(of:)` in ZeroSchemaDecoder.swift:176) is now exercised end-to-end.

Tests/CastTests/ZeroSchemaDecoderTests.swift
```diff
@@ -0,0 +1,246 @@
...
+    var i64: Int64 = 0
+    var u: UInt = 0
+    var u8: UInt8 = 0
+    var u16: UInt16 = 0
+    var u32: UInt32 = 0
+    var u64: UInt64 = 0
+}
+
+private struct RequiredAndOptional: Decodable, Sendable {
+    var a: Int = 0
```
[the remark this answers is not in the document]
jaylann: Verified — no change needed. Swift's synthesized `Decodable.init(from:)` calls `decode(_:forKey:)` for non-Optional stored properties regardless of the property's default value. Defaults only inform the synthesized `init()` (memberwise/Castable), not the `Decodable` init. So `var a: Int = 0` routes through `_KeyedContainer.decode(...)` (which appends to `requiredFields`), and the assertion `info.required == ["a"]` is correctly exercising the required-field path. CI confirms this — the test is green.

Tests/CastTests/ZeroSchemaDecoderTests.swift
```diff
@@ -0,0 +1,246 @@
...
+        #expect(propKeys.contains("x"))
+    }
+
+    @Test("Enum via CastEnum decodes as enumeration with first case zero")
+    func enumViaCastEnum() throws {
+        let info = try ZeroSchemaDecoder.decode(ColorHolder.self)
+        let field = try #require(info.fields.first { $0.name == "color" })
+        #expect(field.kind == .enumeration)
+
+        let zero = try #require(info.zeroInstance as? ColorHolder)
```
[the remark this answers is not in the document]
jaylann: Verified — no change needed. Same Swift-synthesis behavior as the line 27 thread: non-Optional `var color: SwatchColor = .red` routes through `_KeyedContainer.decode<T: Decodable>(...)` which goes to enum schema lookup. The `firstCase(of:)` / enum instantiation path is genuinely exercised — the property default is unrelated to the decode path. Also note: in the same commit (e95823f) the test now additionally asserts the schema content (`Set(json["enum"]) == ["red", "blue"]`) so the enum-instantiation contract has even tighter coverage.

Tests/CastTests/ZeroSchemaDecoderTests.swift
```diff
@@ -0,0 +1,246 @@
+@testable import Cast
+import Collections
```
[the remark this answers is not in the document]
jaylann: Not removed in e95823f — `import Collections` is genuinely required by this test file. The new `WithRangeNumeric` fixture uses `@Castable`, which expands to `OrderedDictionary(dictionaryLiteral: ...)` and `JSONSchema.string(...)` / `.integer(...)` calls. Even though `Sources/Cast/API/Castable.swift` does `@_exported import Collections` / `@_exported import JSONSchema`, `@testable import Cast` does **not** propagate the re-exports — the macro-expanded code fails to compile with `cannot find 'OrderedDictionary' in scope` and `cannot infer contextual base in reference to member 'string'` without the explicit imports in this file. Verified by removing the import locally and observing the build failure. Leaving as-is.

Tests/CastTests/ZeroSchemaDecoderTests.swift
```diff
@@ -0,0 +1,246 @@
...
+        #expect(field.kind == .array(element: .object))
+
+        let data = try JSONEncoder().encode(field.schema)
+        let json = try #require(
+            try JSONSerialization.jsonObject(with: data) as? [String: Any]
+        )
+        let items = try #require(json["items"] as? [String: Any])
+        let props = try #require(items["properties"] as? [String: Any])
+        let propKeys = props.keys.joined(separator: ",")
+        #expect(propKeys.contains("x"))
```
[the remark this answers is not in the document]
jaylann: Fixed in e95823f: replaced both occurrences (lines 158 and 174) with `props.keys.contains(where: { $0.contains("x") })`. The `joined`-then-substring intermediate is gone; the `let propKeys` is also dropped. Note: a literal `props.keys.contains("x")` doesn't work for these `@Castable`-derived nested types — the macro mangles private-type field keys to `__0__x`, so the substring check is intentional, just expressed more directly via `Collection.contains(where:)` instead of joining first.

PR #112, opened 2026-05-01
fix(safety): resolve Sendable capture in iOS background safety (#105)

### LINKED ISSUE #105: Cast doesn't build for iOS — Sendable capture errors in CastModel+Lifecycle.swift

jaylann:
## Summary

Fix three Swift 6 strict-concurrency errors that broke the iOS build of Cast (`enableBackgroundSafety()` in `CastModel+Lifecycle.swift`).

The `_observers.withLock { ... }` closure is `@Sendable`. It was constructing `_LifecycleObservers` inline, which captured three `any NSObjectProtocol` locals (`didEnter`, `willResign`, `memWarn`). `NSObjectProtocol` is not `Sendable`, so the captures were rejected.

Fix: build the `_LifecycleObservers` wrapper before the `withLock` call. The wrapper is `@unchecked Sendable` (already declared in this file), so the closure now captures only a Sendable value. Six lines of churn, no API or behavior change.

## Test plan

- [x] `swift test` passes locally (full suite, including `LifecycleTests`)
- [x] Type-system-driven correctness: a Sendable wrapper captured by a @Sendable closure is well-formed
- [ ] iOS Xcode build verified once #44 demo project is hooked into CI (next step)

Closes #105

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal lifecycle observer initialization for improved code efficiency. No user-facing changes.

### DIFF
--- a/Sources/Cast/API/CastModel+Lifecycle.swift
+++ b/Sources/Cast/API/CastModel+Lifecycle.swift
@@ -109,12 +109,13 @@ public extension CastModel {
                 Memory.clearCache()
             }
 
+            let observers = _LifecycleObservers(
+                didEnterBackground: didEnter,
+                willResignActive: willResign,
+                memoryWarning: memWarn
+            )
             _observers.withLock { dict in
-                dict[key] = _LifecycleObservers(
-                    didEnterBackground: didEnter,
-                    willResignActive: willResign,
-                    memoryWarning: memWarn
-                )
+                dict[key] = observers
             }
         }
 


Sources/Cast/API/CastModel+Lifecycle.swift
```diff
@@ -109,12 +109,13 @@ public extension CastModel {
                 Memory.clearCache()
             }
 
+            let observers = _LifecycleObservers(
+                didEnterBackground: didEnter,
+                willResignActive: willResign,
+                memoryWarning: memWarn
+            )
```
jaylann, line 116: Pre-existing race (not introduced here, not blocking): the `if already { return }` check on line 81 releases the lock before the `addObserver` calls and the final `dict[key] = observers` write. Two concurrent `enableBackgroundSafety()` calls on the same `CastModel` can both pass the check, each register three observers, and the second `withLock` overwrites the first wrapper — orphaning the first set of observers (no `removeObserver` will ever fire for them, because `_observers` only retains the last wrapper).

The right fix is a single `withLock` that does check-then-insert (e.g. register observers first, then `withLock { if dict[key] == nil { dict[key] = observers; return false } else { return true } }`, and on `true` immediately `removeObserver` the three just-registered observers). Worth a follow-up issue, but outside the scope of this PR — the Sendable fix here is correct and minimal.

The Sendable change itself is clean: `_LifecycleObservers: @unchecked Sendable` (line 39) holds because the wrapper is immutable post-init (three `let` props), so the closure capturing only `observers` + `key` is well-formed under Swift 6 strict concurrency. LGTM.

jaylann: Tracked as #114 for follow-up. Resolving here.

PR #113, opened 2026-05-01
feat(api): pluggable ModelSource — directory, bundle, endpoint (#101)

### LINKED ISSUE #101: feat(api): pluggable model sources — local dir, custom endpoint/CDN, bundled resources
## Problem

`CastModel.load(_:)` only accepts a Hugging Face `org/repo` id and always downloads from `huggingface.co` via the default `HubApi.shared`. Users who can't or don't want to hit HF directly currently have to drop down to `MLXLMCommon` and use `init(wrapping:configuration:)` — which contradicts Cast's "one import, one API" promise.

Concrete asks:

1. **Local / pre-downloaded model.** Air-gapped device, CI cache, custom path (`~/Models/llama-3.2-3b-4bit/`).
2. **Custom HF mirror or corporate CDN.** Self-hosted `huggingface.co` mirror, behind-VPN model registry, regional CDN to avoid HF rate limits.
3. **Bundled into the app.** Ship the model inside the `.app` (`Bundle.main.url(forResource:withExtension:)`) so first-launch has no network round-trip and no progress UI is needed.
4. **Pin a specific revision.** Reproducible builds shouldn't silently float to a new HF commit.

`MLXLMCommon.ModelConfiguration` and `Hub.HubApi` already support all of this — Cast just doesn't surface the knobs.

## Proposed API

Add a `ModelSource` enum (or equivalent) and a new overload:

```swift
public enum ModelSource: Sendable {
    /// Hugging Face Hub, default endpoint. Existing behavior.
    case huggingFace(id: String, revision: String? = nil)

    /// Pre-downloaded / on-disk model directory.
    case directory(URL)

    /// Bundled inside the app's resource folder.
    /// Resolves at load time via `bundle.url(forResource: name, withExtension: nil)`.
    case bundle(_ bundle: Bundle = .main, resourceName: String)

    /// Hugging-Face-shaped repo, but downloaded from a custom endpoint
    /// (mirror, internal CDN, proxy). Cache root is also configurable.
    case customEndpoint(id: String, endpoint: URL, downloadBase: URL? = nil, revision: String? = nil)
}

extension CastModel {
    public static func load(
        _ source: ModelSource,
        progress: (@Sendable (Progress) -> Void)? = nil
    ) async throws -> CastModel
}
```

Keep the existing string overload (`load(_ modelId: String, …)`) as a thin wrapper over `.huggingFace(id:)` for source compatibility.

### Mapping to underlying types

| `ModelSource`     | Underlying `ModelConfiguration` / `HubApi`                                    |
|-------------------|--------------------------------------------------------------------------------|
| `.huggingFace`    | `ModelConfiguration(id: id, revision: revision)` + `HubApi.shared`             |
| `.directory`      | `ModelConfiguration(directory: url)` + `HubApi.shared` (only for tokenizer fetch if missing locally — caller is responsible for completeness) |
| `.bundle`         | Resolve URL via `bundle.url(forResource:)`, then same as `.directory`         |
| `.customEndpoint` | `ModelConfiguration(id: id, revision: revision)` + `HubApi(downloadBase: …, endpoint: …)` |

## Open questions

- Should `.bundle` validate the resource directory shape (model.safetensors + tokenizer.json + config.json) up front, or fail late with the underlying MLX error?
- Do we want a sibling `CastModel.preflight(_:)` that resolves a `ModelSource` to a local path without loading weights, useful for "is this model installed?" UIs?
- For `.customEndpoint`, expose a custom on-disk cache root too (`HubApi(downloadBase: …)`) — see issue followup if not in v1.

## Acceptance criteria

- [ ] `ModelSource` enum lands in `Sources/Cast/API/`, public + `Sendable`.
- [ ] `CastModel.load(_ source: ModelSource, progress:)` overload works for all four cases.
- [ ] String overload preserved; existing call sites compile unchanged.
- [ ] Examples: new `Examples/Sources/CustomModelSource/main.swift` covering at least `.directory` and `.customEndpoint`. (Bundled requires an app target — defer or use a placeholder.)
- [ ] DocC article + `Cast.md` topic entry (per `Examples/CLAUDE.md`).
- [ ] Tests: schema-level tests for `ModelSource` mapping; runtime load tests gated with `.requiresMetal`.
- [ ] README install/usage snippet updated to mention non-HF options.

## Out of scope (follow-ups)

- Auto-discovery of bundled models at app launch.
- A registry/manifest format for distributing model bundles separately from app binaries.
- Authenticated endpoints (HF token, corporate auth headers) — track separately if needed.

## Context

- `CastModel.load(_:)` lives at `Sources/Cast/API/CastModel.swift:68-79`.
- `init(wrapping:configuration:)` (current escape hatch) at `Sources/Cast/API/CastModel.swift:53`.
- Vendored `Sources/MLXStructured/GrammarMatcherFactory.swift:16-21` already branches on `.id` vs `.directory` `ModelConfiguration` cases — no plumbing changes needed below the API.
- `Examples/Sources/CallerManagedLoading/main.swift` is the current "workaround" consumers use; the new API would replace that pattern.

jaylann:
## Summary

Adds a `ModelSource` enum and `CastModel.load(_ source: ModelSource, progress:)` overload covering four use cases users currently escape-hatch via `init(wrapping:configuration:)`:

- `.huggingFace(id:revision:)` — default HF Hub, with optional revision pinning
- `.directory(URL)` — pre-downloaded / air-gapped models
- `.bundle(_:resourceName:)` — model shipped inside the app
- `.customEndpoint(id:endpoint:downloadBase:revision:)` — corporate HF mirror / self-hosted CDN

Existing `CastModel.load(_ modelId: String, ...)` is preserved as a thin wrapper over `.huggingFace(id:)`, so all existing call sites compile unchanged. New `CastError.modelNotFound` for missing bundle resources.

## Test plan

- [x] `swift test` passes (new `ModelSourceTests` + existing suites)
- [x] `cd Examples && swift build` passes (new `CustomModelSource` example)
- [x] DocC article auto-generated via `scripts/generate-example-docs.sh`
- [x] README mentions non-HF options
- [ ] Manual `.directory` smoke-test on local cached model (requires Metal — local-only)

Closes #101

## Summary by CodeRabbit

* **New Features**
  - Added a ModelSource API and a CastModel.load overload to support Hugging Face, local directories, app bundles, and custom mirror endpoints
  - Added a new example executable demonstrating local and mirrored-endpoint model loading

* **Documentation**
  - README guidance updated; new example docs and a repository learning note added

* **Bug Fixes**
  - Added CastError.modelNotFound with clearer message

* **Tests**
  - Reworked CastError tests and added ModelSource resolution tests


Tests/CastTests/ModelSourceTests.swift
```diff
@@ -0,0 +1,109 @@
...
+        let endpoint = try #require(URL(string: "https://example.com"))
+        let source = ModelSource.customEndpoint(id: "x/y", endpoint: endpoint)
+        let (configuration, _) = try source.resolved()
+        guard case let .id(_, revision) = configuration.id else {
+            Issue.record("Expected .id identifier, got \(configuration.id)")
+            return
+        }
+        #expect(revision == "main")
+    }
+}
```
jaylann, line 136: Test gap: the `customEndpoint` cases only assert `id` and `revision` propagated into `ModelConfiguration`, but the **whole point** of `.customEndpoint` is that the resolved `HubApi` carries the supplied endpoint (and optional `downloadBase`). `HubApi.endpoint` is `internal var` — not directly inspectable from a test target — but the resolution can still be exercised by:

- Constructing two sources (one `.huggingFace`, one `.customEndpoint`) and asserting the returned `HubApi` instances are not `===` / not equal to `.shared` for the custom case (`.shared` is used for `.huggingFace`/`.directory`/`.bundle`).
- Or making `HubApi` introspectable via `@testable import Hub` and asserting `endpoint`/`downloadBase` directly.

As-is, swapping `HubApi(downloadBase: downloadBase, endpoint: endpoint.absoluteString)` for `HubApi.shared` in `resolved()` would silently pass these tests — i.e. the regression-detection value of the `customEndpoint` tests is near zero. Add at minimum an identity-vs-shared check.

Sources/Cast/API/ModelSource.swift
```diff
@@ -0,0 +1,51 @@
...
+        switch self {
+        case let .huggingFace(id, revision):
+            let revision = revision ?? "main"
+            return (ModelConfiguration(id: id, revision: revision), .shared)
+
+        case let .directory(url):
+            return (ModelConfiguration(directory: url), .shared)
+
+        case let .bundle(bundle, resourceName):
+            guard let url = bundle.url(forResource: resourceName, withExtension: nil) else {
```
jaylann: `bundle.url(forResource: resourceName, withExtension: nil)` is ambiguous when there's a same-named file with an extension alongside a same-named directory resource (or two same-named files with different extensions). Apple's docs: passing `nil` returns the first file encountered with that base name, regardless of extension — so a stray `model.json` next to a `model/` directory could win and resolve to the wrong URL.

For a model **directory** resource, the intent is "no extension". Prefer `withExtension: ""` (Apple's documented sentinel for "no extension") or use the directory-aware overload `url(forResource:withExtension:subdirectory:)` with explicit `subdirectory: nil`. As-is the current call works for the happy path but is fragile against innocuous bundle layouts.

CLAUDE.md
```diff
@@ -129,3 +129,5 @@ docs/
...
+- `import Hub` (from `huggingface/swift-transformers`) is **not** transitively re-exported through `MLXLMCommon`/`MLXStructured` — adding `HubApi` directly to a `Sources/Cast/` file requires `.product(name: "Hub", package: "swift-transformers")` on the `Cast` target in `Package.swift`. `Hub` is already in `MLXStructured`'s deps but downstream targets pulling in `Cast` don't see it without the explicit edge. Surfaced when adding `ModelSource.customEndpoint` (#101).
+- Git worktrees created off this repo do **not** inherit submodule checkouts — `git submodule update --init --recursive` must be re-run inside the worktree before the first `swift build`, otherwise SPM emits `Invalid Exclude '…/CMLXStructured/xgrammar/…': File not found` warnings and the build fails to find xgrammar headers. CLAUDE.md mentions this for fresh clones; same applies to `git worktree add` outputs.
```
jaylann: This learning is **already** captured on `stage` — the existing `Parallel-PR git worktrees can't live in ../Cast.worktrees/...` line was extended in a prior commit to include: *"After `worktree add`, run `git submodule update --init --recursive` from inside the worktree before `swift build`/`swift test` — Cast vendors `Sources/CMLXStructured/xgrammar` as a submodule, which `git worktree add` does NOT initialize, and SPM fails to resolve `CMLXStructured` without it."*

The new appended line duplicates that guidance with slightly different framing. Suggest dropping this line and (if anything new) augmenting the existing one in place — otherwise the file accumulates near-duplicate learnings that are easy to miss when scanning.

Sources/Cast/API/CastModel.swift
```diff
@@ -69,8 +73,40 @@ public final class CastModel: Sendable {
...
+    public static func load(
+        _ source: ModelSource,
+        progress: (@Sendable (Progress) -> Void)? = nil
+    ) async throws -> CastModel {
+        let (configuration, hub) = try source.resolved()
         let container = try await LLMModelFactory.shared.loadContainer(
+            hub: hub,
             configuration: configuration
         ) { prog in
             progress?(prog)
```
jaylann: The doc says the progress callback is "only meaningful for `.huggingFace` and `.customEndpoint`" — but the implementation silently passes `progress` into `LLMModelFactory.shared.loadContainer` for every `ModelSource`, including `.directory`/`.bundle` where there is nothing to download.

For `.directory`/`.bundle`, callers will register a callback and never see progress events fire (no events) and have no programmatic way to detect whether progress is supported for the source they passed. A few options, in increasing strictness:

1. (Lightest) Strengthen the doc — note the callback is a **no-op** for local sources, not just "not meaningful".
2. Skip the callback for local sources entirely (passing `nil` through, or no closure).
3. Move progress to the `.huggingFace` / `.customEndpoint` cases as an associated value, so the type system enforces it.

(1) is probably enough for v1; flagging because the current wording is easy to misread as "you'll get a single 'done' event".

Sources/Cast/API/CastError.swift
```diff
@@ -22,6 +22,8 @@ public enum CastError: LocalizedError, Sendable {
     /// or `abortInFlight()`). `partialOutput` is the bytes generated up to
     /// the cancel point.
     case cancelled(partialOutput: String?)
+    /// A model source could not be resolved (e.g. missing bundle resource).
+    case modelNotFound(String)
```
jaylann: `CastError.modelNotFound(String)` is a fine case name, but the doc comment scopes it narrowly to bundle resources (`"e.g. missing bundle resource"`). Today it's only thrown for `.bundle`, but `.directory` currently doesn't validate its URL exists at all — `ModelConfiguration(directory: url)` accepts any URL and the failure surfaces later as an MLX/file-IO error, wrapped or not.

Consider either:
- Validating `.directory`'s URL up front in `resolved()` and throwing `.modelNotFound` for symmetry (so users get a single, well-typed error for "the path you gave me doesn't exist"), **or**
- Renaming/scoping the case to `.bundleResourceNotFound` so the case name matches its single call site.

As-is the case is broader than what it documents but narrower than what its name implies, which is the worst of both worlds for API stability once it ships.

Sources/Cast/API/CastModel.swift
```diff
@@ -69,8 +73,40 @@ public final class CastModel: Sendable {
...
+    /// // Local pre-downloaded model
+    /// let model = try await CastModel.load(
+    ///     .directory(URL(fileURLWithPath: "/Users/me/Models/llama-3.2-3b-4bit"))
+    /// )
+    ///
+    /// // Corporate HF mirror
+    /// let model = try await CastModel.load(
+    ///     .customEndpoint(
+    ///         id: "internal/llama-3.2-3b-4bit",
+    ///         endpoint: URL(string: "https://hf-mirror.corp.example.com")!,
```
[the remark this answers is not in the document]
jaylann: Applied in 8d0f69b — replaced `URL(string: ...)\!` with a `guard let endpoint = URL(string: ...) else { ... }` form in the doc example.

Sources/Cast/API/ModelSource.swift
```diff
@@ -0,0 +1,51 @@
+import Foundation
+import Hub
+import MLXLMCommon
+
+/// Where a model's weights and tokenizer live. Use with
+/// ``CastModel/load(_:progress:)-(_,_)``.
```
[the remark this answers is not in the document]
jaylann: Applied in 8d0f69b — link now points to the `(ModelSource,_)` overload disambiguation.

Tests/CastTests/ModelSourceTests.swift
jaylann: Applied in 8d0f69b — added behavioral assertions for `.customEndpoint` that the resolved `HubApi` carries the supplied `endpoint` and `downloadBase`, and is a fresh instance distinct from `HubApi.shared`. Required `@testable import Hub` and a Hub dep on the CastTests target.

Sources/Cast/API/ModelSource.swift
jaylann: Applied in 8d0f69b — switched to `withExtension: ""` (Apple's documented sentinel for no extension) so directory resources match unambiguously.

CLAUDE.md
jaylann: Resolved during the merge from `stage` (42e6c9a) — the duplicate worktree-submodule learning is gone; only the new `import Hub` non-re-export learning remains, which is unique.

Sources/Cast/API/CastModel.swift
jaylann: Applied in 8d0f69b — strengthened the doc to explicitly note the callback is a no-op for `.directory` and `.bundle` sources (no download, no events fire). Implementation unchanged.

Sources/Cast/API/CastError.swift
jaylann: Applied in 8d0f69b — broadened the `.modelNotFound` doc to cover both bundle resources and future `.directory` existence validation. Case name and `String` payload unchanged.

Tests/CastTests/ModelSourceTests.swift
```diff
@@ -0,0 +1,140 @@
...
+            Issue.record("Expected resolved() to throw")
+        } catch let error as CastError {
+            guard case .modelNotFound = error else {
+                Issue.record("Expected .modelNotFound, got \(error)")
+                return
+            }
+        } catch {
+            Issue.record("Expected CastError, got \(error)")
+        }
+    }
```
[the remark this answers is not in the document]
jaylann: Applied in d98de1d — collapsed to a single typed `#expect { try source.resolved() } throws: { error in ... }` form per the suggestion, dropping the redundant manual do-catch.

Examples/Package.swift
```diff
@@ -47,6 +47,10 @@ let package = Package(
                 .product(name: "MLXLMCommon", package: "mlx-swift-lm")
             ]
         ),
+        .executableTarget(
+            name: "CustomModelSource",
+            dependencies: [.product(name: "Cast", package: "Cast")]
+        ),
```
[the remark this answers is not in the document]
jaylann: Applied in 068da77 — added `path: "Sources/CustomModelSource"` per the Examples/CLAUDE.md guideline.

